PR #1, opened 2016-01-05
Add trailing backslash to fix "not a directory" error

fmaker:
Running this Dockerfile on Fedora 22 throws a "not a directory" error when it reaches "ADD cert/cert.pem /cert". Adding a backslash to the path resolves the issue.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -9,8 +9,8 @@ ADD mattermost /etc/nginx/sites-available/
 ADD docker-entry.sh /
 
 RUN mkdir /cert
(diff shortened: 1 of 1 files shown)


npcode:
Good point! Thank you for the fix.

I guess this is the reason of the error:

> If `<dest>` does not end with a trailing slash, it will be considered a regular file and the contents of `<src>` will be written at `<dest>`.
>
> -- https://docs.docker.com/engine/reference/builder/

fmaker:
Thank you for the script! Very helpful.

PR #4, opened 2016-03-20
Proposed update to v2.1.0

it33:


npcode:
Thanks a lot, ian! Now we can download the latest release from releases.mattermost.com!

PR #8, opened 2016-04-16
Allow db host, name user and password to be configurable

xcompass:
Those parameters can be configured through environment variables now.


npcode:
Confirmed. Thanks for your contribution!

I think we may add the default environments in docker-compose.yml so that users easily know how to modify them.

PR #18, opened 2016-05-07
Fix #15, WAL in archive mode when backup is off

### LINKED ISSUE #15: [BUG] Lack of AWS S3/wal-e backup leads to full disks
Hi again guys,

We've been running this instance for a few days now and in this last hour I was struck with a very nice error of database, the db container died out of lack of space in disk. the `volumes` folder had 53GB 😮 .

Turns out that, if wal-e fails to backup, the postgresql db will start saving various WAL files in a folder called `pg_xlog` (this was the 53GB monster).

I followed this [post](http://blog.endpoint.com/2014/09/pgxlog-disk-space-problem-on-postgres.html) to recover everything and restart my mattermost server without any issue 😄 

All safe guys!
Since this is planned to use in production, this could be a issue...
Thank you all,
Psidium

xcompass:
WAL logs in archive mode will consume a lot more spaces and not useful
when backup is off. This fix add ability to switch WAL logs mode based
on the backup switch.

### DIFF
--- a/db/docker-entrypoint1.sh
+++ b/db/docker-entrypoint1.sh
@@ -1,5 +1,17 @@
 #!/bin/bash
 
+# if wal backup is not enabled, use minimal wal logging to reduce disk space
+: ${WAL_LEVEL:=minimal}
+: ${ARCHIVE_MODE:=off}
+# PGDATA is defined in upstream postgres dockerfile
+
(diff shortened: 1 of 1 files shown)


npcode:
Thanks for the PR, @xcompass !

PR #19, opened 2016-05-07
Use $PGDATA env var to setup wal-e

xcompass:
$PGDATA is set in upstream postgres dockerfile and should be used in
downstream in case the postgres data directory is changed in the future.

This PR depends on #18


npcode:
Could you please rebase this on master?

xcompass:
Done

npcode:
Thanks!

PR #23, opened 2016-05-20
Add .gitattributes file to ensure correct line endings on Windows

### LINKED ISSUE #20: How to fix `ERROR: for db Container command '/docker-entrypoint1.sh' could not be invoked` on Windows
I thought I'd try out this Docker setup, but after all the images are built, I get this message:

```
ERROR: for db Container command '/docker-entrypoint1.sh' could not be invoked.
```

I don't get any debugging messages, but a bit of sleuthing discovered that this is caused by a couple Docker-and-Git-on-Windows issues:

* Windows doesn't support the executable bit, and
* Windows Git clients are usually set up with `autocrlf`, which adds a `^M` to the end of the `#!/bin/bash` line in `db/docker-entrypoint1.sh`.

So, for Windows users I'd recommend the following:

* In the repo's directory, run `git config core.autocrlf false` and checkout the code again (using `git reset --hard HEAD` or whatever you prefer).
* In `db/docker-entrypoint1.sh`, add the line `RUN chmod +x /docker-entrypoint1.sh` after the line `COPY docker-entrypoint1.sh /`.

These two things allowed me to run Mattermost without troubles on Windows!

andymac4182:
I needed to add this to fix an issue with docker-compose and Windows. 

This should help resolve #20


.gitattributes
```diff
@@ -0,0 +1,4 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
+###############################################################################
+* text eol=lf
```
npcode: I think `*.sh eol=lf` is better because only shell scripts matter. What do you think about it?

andymac4182: The reason I did all is linux doesn't require .sh for executable scripts. 

The other reason is a personal preference that the repo uses 1 line ending for all files to make it simple to know what should be used.

npcode: According to the [gitattributes man page](https://git-scm.com/docs/gitattributes), `eol=lf` enables end-of-line normalization **without any content checks**. I am afraid that it may ruin non-text files which might be added in future. Actually I already have binary files in my custom branches.

esethna:
Hi there, @npcode @andymac4182 just checking in on if there are next steps for this PR?

npcode:
@esethna I hope any safer way. e.g.
```
*.sh text eol=lf
```

pichouk:
Hi @andymac4182, is there anything new for this PR. I guess it could be a good idea but we maybe should use a better regex thant `*` to match only files that matter.

andymac4182:
Sorry I don't have anything extra to add. I unfortunately don't have the time to check it out at the moment.

PR #32, opened 2016-06-13
Removed check if /mattermost/config/config.json already exists before…

Tobatus:
… using config.template and environment values to create new config.json.

When Dockerfile install mattermost it will always write /mattermost/config/config.json file, therefore using environment values to create new config.json never happens if there is this file check.

Also if one whishes to restart docker with new database value there is good reason to always create new config file from environment values.

### DIFF
--- a/app/docker-entry.sh
+++ b/app/docker-entry.sh
@@ -6,18 +6,14 @@ MM_USERNAME=${MM_USERNAME:-mmuser}
 MM_PASSWORD=${MM_PASSWORD:-mmuser_password}
 MM_DBNAME=${MM_DBNAME:-mattermost}
 echo -ne "Configure database connection..."
-if [ ! -f $config ]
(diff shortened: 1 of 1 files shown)


npcode:
> When Dockerfile install mattermost it will always write /mattermost/config/config.json file, therefore using environment values to create new config.json never happens if there is this file check.

No, if in the host ./volumes/app/mattermost/config/config.json exists mattermost reuse it and that is my intention.

Tobatus:
Ok. So app-docker only works when attaching volume when running it? Isn't that bit complicated? How about just delete extra /mattermost/config/config.json in Dockerfile after extracting .tar.gz? So intended usage with volumes still works, but it would be possible to run without.

npcode:
@Tobatus Good point. I'll try to fix it.

PR #62, opened 2016-11-15
Update docker-compose file to v2 format

Andrey9kin:

PR #72, opened 2016-12-17
Bump docker-compose.yml version to 2.1

banzo:
I needed to use the build arg feature, which is only available in recent docker-compose versions.


banzo:
Duplicates #62 sorry

Issue #84, opened 2017-01-27
Dead link on README


Niceplace:
At the end of the README, the following line contains a link "For the server configurations, see prod-ubuntu.rst of mattermost.".

The link is "prod-ubuntu.rst" which points to https://github.com/mattermost/docs/blob/master/source/install/prod-ubuntu.rst. 

This link seems to be broken.

jasonblais:
Thanks @Niceplace for the note, I've submitted a pull request for it.

The Ubuntu guides can be found here, looks like the link got broken due to the recent installation guide updates: https://docs.mattermost.com/install/install-ubuntu-1404.html

PR #85, opened 2017-01-27
Update link to install guide

jasonblais:
Addresses #84


jasonblais:
@xcompass Comes to mind - I'm not sure if we want to have references to our other install guides?

We currently have four separate guides

image

xcompass:
@jasonblais How about adding a link I add a link to the above page in the readme in case someone is not using docker?

jasonblais:
Yeah, that would work. The link to that page is: https://docs.mattermost.com/guides/administrator.html#installing-mattermost

xcompass:
Thanks. Done.

PR #98, opened 2017-02-20
simplify reverse proxy

nikosch86:
- make use of mainline nginx image
- switch ssl/non-ssl config file based on presence of cert/key
- hardcode app port since it is hardcoded in the app container as well

this will allow following changes in the mainline nginx images rather than building our own nginx


docker-compose.yml
```diff
@@ -40,7 +40,3 @@ services:
     volumes:
       # This directory must have cert files
       - ./volumes/web/cert:/cert:ro
-      - /etc/localtime:/etc/localtime:ro
```
xcompass: Any reason to remove this?

docker-compose.yml
```diff
@@ -40,7 +40,3 @@ services:
...
-    environment:
-      - MATTERMOST_ENABLE_SSL=false
-      - PLATFORM_PORT_80_TCP_PORT=80
```
xcompass: Could you also remove the reference in readme file?

docker-compose.yml
nikosch86: i added this line again, i think this change will spark a discussion that does not belong in this PR

docker-compose-v1.yml
```diff
@@ -39,7 +39,3 @@ web:
   volumes:
       # This directory must have cert files
     - ./volumes/web/cert:/cert:ro
-    - /etc/localtime:/etc/localtime:ro
```
xcompass: Could you ass this back? Otherwise, this looks great.

nikosch86: i did
https://github.com/mattermost/mattermost-docker/pull/98/commits/032b5412d67f249d28ac3b6b1671aa093e2386eb

xcompass: sorry for the typo. Auto correct... I saw you added it back in docker-compose.yml. This is removed from docker-compose-v1.yml. Thanks.

xcompass:
Looks good. Thanks!

PR #100, opened 2017-02-20
make use of postgres image db init script, use their variable naming …

nikosch86:
…scheme, make setup-wale executable so it can actually run


docker-compose-v1.yml
```diff
@@ -6,9 +6,9 @@ db:
     - /etc/localtime:/etc/localtime:ro
   # uncomment the following to enable backup
   environment:
-    - MM_USERNAME=mmuser
-    - MM_PASSWORD=mmuser_password
-    - MM_DBNAME=mattermost
+    - POSTGRES_USER=mmuser
```
xcompass: Could you add a comment let people know those values has to be the same as environment variables in app service?

xcompass:
Could also you update the MM_XXXXXX references in readme.md for database section?

Any idea why test failed?

nikosch86:
since you are using docker run in your tests instead of the compose file, the database is not actually created, before these changes that was the case because the make db script hardcoded it

Issue #124, opened 2017-05-17
DB container fails to build


tumbl3w33d:
When I build the db container using [this Dockerfile](https://github.com/mattermost/mattermost-docker/blob/master/db/Dockerfile) from current master then the `docker build` fails building the cryptography module for python due to missing openssl and ffi headers.

Fix:
Add installation of packages `libssl-dev` `libffi-dev`.

thomasnaudin:
Same problem here, @tumbl3w33d solution worked for me. Thanks :)

ipsecguy:
Same here:

`creating build/temp.linux-x86_64-2.7/build/temp.linux-x86_64-2.7

    x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fno-strict-aliasing -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security -fPIC -I/usr/include/python2.7 -c build/temp.linux-x86_64-2.7/_openssl.c -o build/temp.linux-x86_64-2.7/build/temp.linux-x86_64-2.7/_openssl.o
    build/temp.linux-x86_64-2.7/_openssl.c:434:30: fatal error: openssl/opensslv.h: No such file or directory

     #include <openssl/opensslv.h>

                                  ^
    compilation terminated.

    error: command 'x86_64-linux-gnu-gcc' failed with exit status `

pichouk:
Should be now fixed by #129

tumbl3w33d:
Good stuff, thanks.

PR #126, opened 2017-05-18
Update database dockerfile

pichouk:
Should fix issue https://github.com/mattermost/mattermost-docker/issues/124


ipsecguy:
Does not look like it is applied, Dockerfile for DB is still 3 month since last update.

pichouk:
Closing because #129 seems to fix it in a better way.

PR #131, opened 2017-06-01
Improve app Dockerfile

pichouk:
Some cleanup on the app Dockerfile to made them lighter and prettier


jasonblais:
This reminds me, Mattermost now supports overriding config.json settings with environment variables in case that's of interest @pichouk

pichouk:
Interesting. It just noticed that the `config.template.json` file is overriding the default `/mattermost/config/config.json` file which, have now a lot more keys.
I will see if we can use environment variables instead, to stay up-to-date with original `config.json` file.

pichouk:
I'm just afraid that it result in a **loooooooooong** list of `ENV` instructions on the Dockerfile. It might have a lot of setting to override.

jasonblais:
Yeah, that's a good point. Something to think about I suppose :)

pichouk:
I just rebase the `config.json` file from the main Mattermost project and merge it with our configuration.
I also tried to use only environment variables to override default settings : it is quite ugly, if anyone have a better idea on how to do it I'm interesting.
Maybe we should keep the custom `config.template.json` file but it need to be rebase regularly...

I will anyway try this on my production instance to see if it work.

pichouk:
I'm not sure why tests failed, but it build and run on my Debian server...

xcompass:
@pichouk @jasonblais Ideally, I think upstream mattermost should support environment override for the configs.

app/Dockerfile
```diff
@@ -1,18 +1,39 @@
...
+    MM_EMAILSETTINGS_SENDEMAILNOTIFICATIONS=false \
+    MM_EMAILSETTINGS_FEEDBACKEMAIL="" \
+    MM_EMAILSETTINGS_SMTPSERVER="" \
+    MM_EMAILSETTINGS_SMTPPORT="" \
+    MM_EMAILSETTINGS_INVITESALT="bjlSR4QqkXFBr7TP4oDzlfZmcNuH9YoS" \
+    MM_RATELIMITSETTINGS_ENABLE=true \
+    MM_SQLSETTINGS_DRIVERNAME="postgres"
+
+# Install some needed packages
+RUN apt-get update; \
```
xcompass: I think the best practice for joining the command is using "&&", which will cause the whole command fail if one of the failed. Otherwise, I think if the last command successfully completed, the whole command is considered successful.

app/docker-entry.sh
```diff
@@ -21,21 +21,15 @@ if [ "$1" = 'platform' ]; then
...
-        sed -Ei "s/MM_PASSWORD/$MM_PASSWORD/" $MM_CONFIG
-        sed -Ei "s/MM_DBNAME/$MM_DBNAME/" $MM_CONFIG
-        echo OK
-    else
-        echo SKIP
+        cp /config.json.save $MM_CONFIG
     fi
 
+    echo -ne "Configure database connection..."
+    export MM_SQLSETTINGS_DATASOURCE="postgres://$MM_USERNAME:$MM_PASSWORD@$DB_HOST:$DB_PORT/$MM_DBNAME?sslmode=disable&connect_timeout=10"
```
xcompass: Is this env var supported by MM? I can't find any docs.

app/Dockerfile
pichouk: You're right, it is better with `&&`. I'll fix it.

app/docker-entry.sh
pichouk: I guess it should be since all vars can be override : https://docs.mattermost.com/administration/config-settings.html

pichouk:
@xcompass I made some modifications and it works now.
Since it introduce a "breaking" change on the database port number, I'll prefer to wait for your review :smile:

xcompass:
Thanks @pichouk This looks good!

PR #133, opened 2017-06-02
Move AWS config

pichouk:
Just moving AWS readme and conf to to `contrib/aws` folder (since there is a `contrib/kubernetes` folder too).
Also improve a little the main README file.


pichouk:
I let PR open for few days in case someone disagree on that organization ;).

Issue #109, opened 2017-03-06
Alpine variant?


ptman:
Ubuntu 14.04 based image is a bit old and heavy. Since mattermost is a go project, it should be able to compile it using CGO_ENABLED=0 and get a totally static binary (`ldd $binary` gives "not a dynamic executable"). And it could be based on alpine, which is very slim compared to ubuntu. Any plans to do that?

cjkoenig:
Hi @ptman, regarding alpine based images, I updated the alpine branch recently and created PR #92. Right now, I do not know the status and if its ever get commited (travis is OK). However, feel free to try it :-)

tejasbubane:
Can we upgrade only `nginx` and `postgres` containers to alpine, until we test the app container for other issues?

ptman:
I've had good success with nginx:alpine and postgres:alpine

giabao:
@ptman 
> I've had good success with nginx:alpine and postgres:alpine
Could you share the code

pichouk:
@ptman If you have something working, I invite you to submit PR :smile:

cybershambles:
This looks fine for me.. but it's also been a few months

I'll test and confirm these PRs and hopefully assist with the merge

cybershambles:
We do need to consider existing instances and plan an upgrade path

unless it's `docker-compose down; docker-compose rm; docker-compose build; docker-complose up -d`

pichouk:
@cybershambles There is on opened PR https://github.com/mattermost/mattermost-docker/pull/135
But this change needs to be perform step by step, with some tests. The PR is stuck for now because I was not able to update my existing database with an Alpine-based one.

pichouk:
As #208 is merged, we are now using Alpine based image :)

PR #92, opened 2017-02-06
Rebase alpine branch and update base images

cjkoenig:
I updated the alpine branch to more recent versions but now I realize that this branch is stale for a while...is there a chance to get alpine based images into the master branch? => They are much smaller than the ubuntu based ones.

PR #135, opened 2017-06-06
Convert db and web containers to use alpine images

tejasbubane:
Ref: #109, #92

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -1,13 +1,12 @@
-FROM postgres:9.4
+FROM postgres:9.4-alpine
 
-RUN apt-get update \
-    && apt-get install -y python-dev libffi-dev libssl-dev lzop pv daemontools curl build-essential \
-    && curl --silent --show-error --retry 5 https://bootstrap.pypa.io/get-pip.py | python \
+RUN apk update \
+    && apk add --no-cache python python-dev ca-certificates gcc linux-headers musl-dev \
+       libffi libffi-dev openssl openssl-dev bash \
+    && apk add py-pip \
+    && pip install --upgrade pip \
     && pip install 'wal-e<1.0.0' \
-    && apt-get remove -y build-essential python-dev \
-    && apt-get autoremove -y \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    && apk del python-dev gcc linux-headers musl-dev libffi-dev openssl-dev
 
 COPY setup-wale.sh /docker-entrypoint-initdb.d/
 COPY docker-entrypoint1.sh /
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:mainline
+FROM nginx:alpine
 
 RUN rm /etc/nginx/conf.d/default.conf
 COPY ./mattermost /etc/nginx/sites-available/
--- a/web/docker-entry.sh
+++ b/web/docker-entry.sh
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/bin/sh
+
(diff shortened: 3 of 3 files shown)


tejasbubane:
Not sure why the build fails. Am I missing something? cc @pichouk

pichouk:
I think you should change the shebang on `web/docker-entry.sh` and `db/docker-entrypoint1.sh` to use `/bin/sh` instead of `/bin/bash` to be use in Alpine.

tejasbubane:
@pichouk Still the same.

pichouk:
It seems you have a syntax error on the `docker-entrypoint1.sh` file of database :
```
/docker-entrypoint1.sh: line 27: syntax error: unexpected "(" (expecting "fi")
```
This is because it use `sh` and not `bash`. I forgot that it could lead to syntax error, this file need to be refactor to work with `sh`.

tejasbubane:
@pichouk I added `bash` package to alpine in order to keep it compatible with the script. I tried with `sh` and `ash`(which comes with alpine) but they don't see to support arrays which are used in the script. 

PS: `bash` is added only in db container, not in web since web works fine with `sh`.

pichouk:
@tejasbubane Looks good to me if there is no other solution for now.
I will test the database container for few days on a production environment to  be sure we are safe with this :)

@xcompass I'll wait for your review/opinion on this PR before merging anyway :smile:

xcompass:
@pichouk When you test the db container, could you test the WAL backup to S3 as well if possible? Thanks.

pichouk:
@xcompass I never used AWS S3, I'll see if I could test this but I'm not sure.

@tejasbubane  I tried the database image on production and I run to the following issue :
```
AWS_ACCESS_KEY_ID is required for Wal-E but not set. Skipping Wal-E setup.
LOG:  invalid value for parameter "log_timezone": "localtime"
LOG:  invalid value for parameter "TimeZone": "localtime"
FATAL:  configuration file "/var/lib/postgresql/data/postgresql.conf" contains errors
AWS_ACCESS_KEY_ID is required for Wal-E but not set. Skipping Wal-E setup.
```
I rollback because, you know, it is production :smile:. I will made some tests locally this week-end to (try to) figure out what's wrong.

I didn't test the Web container since my production infra use another reverse proxy (FYI [Traefik](https://traefik.io/)). I will test locally.

pichouk:
@tejasbubane Did you manage to made the database container works on an existing Mattermost instance ?
I try to run this on my already setup server and I run to
```
FATAL:  configuration file "/var/lib/postgresql/data/postgresql.conf" contains errors
```

It works if I build a new server from scratch, but not with an existing one. I never add any modifications to my `postgresql.conf` file, it is the default one.

haivala:
What's the situation with this?

pichouk:
@haivala This is stale for now. If someone can found times to really tests this and be sure that it will not break anyone, we can then move on.

mritd:
I think we should remove the `--no-cache` option when using the` apk add` or `apk update` command multiple times - because it avoids downloading the index file repeatedly and accelerates the build; 

Use `rm -rf /var/cache/apk/*` clear the index at the completion of the build; 

Compile tools should use `---virtual .build-deps`  to facilitate post-cleaning, like [nignx Dockerfile](https://github.com/nginxinc/docker-nginx/blob/1d2e2ccae2f6e478f628f4091d8a5c36a122a157/mainline/alpine/Dockerfile#L55)

mritd:
The web dockerfile should avoid using the alpine tag without the version number; because it always points to the latest version of build; the docker will cache image, and the image with no version number thereafter will never be updated.

pichouk:
Thanks @mritd for your suggestions.

I'm running out of time now to work on this, but I'll keep what you said in mind when we will go forward on Alpine version. If you want, you can also submit a PR based on the current one.

pichouk:
Closing since we work on it with PR #208 #207 #198

PR #137, opened 2017-06-08
Do not use env variables to override APP conf. (fix #136)

### LINKED ISSUE #136: Newer mattermost-prod-app containers appear to not save config changes
Hi,

It appears that newer versions of this container are not able to save config changes (i.e. things changed in the `System Console`).  I noticed this when, after updating yesterday, my Mattermost install went into `Preview Mode`.  I went into the `System Console` to re-enable sending emails, clicked save, and it immediately went back to `false` again.

I tried to view more detailed logs, and tried to change the log level to `DEBUG`, but just like the email setting, it refused to save my change and reverted back to its default setting of `INFO`.

I also tried to manually edit the `config.json` file to enable email notifications, but that didn't seem to work.  The entry in the file stayed set to `true`, but in the `System Console` web UI, it was still set to `false`.

I then reverted back to container id `91905885c2b2` (https://hub.docker.com/r/mattermost/mattermost-prod-app/builds/bfebjwpehhhjdayqbdsxipm/), and everything works as expected again.  At the time of this writing, there appears to be two newer containers, and I believe I tried both of them, and they both seemed to exhibit the same issue.

Thanks!

pichouk:
PR #131 introduced a **big** issue which cause Mattermost app to not be configurable (as described on #136).
It seems that using environment variables to modify the default Mattermost configuration (as described [here](https://docs.mattermost.com/administration/config-settings.html)) is not a really good idea. Instead, Mattermost app should have the following behavior :
- if a `config.json` file already exists (using volumes mount), do not modify configuration. Only override database connection (since it **should** be done by passing environment vars to app container)
- if there is no `config.json` use the default one and change some vars using `jq`

I run test locally and it respect this behavior. Looks good to me.
@xcompass @br00tal

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -4,25 +4,11 @@ FROM ubuntu:14.04
 ENV PATH="/mattermost/bin:${PATH}"
 ENV MM_VERSION=3.9.0
 
-# Override default config
-ENV MM_SERVICESETTINGS_LISTENADDRESS=":80" \
-    MM_LOGSETTINGS_ENABLECONSOLE=false \
-    MM_LOGSETTINGS_CONSOLELEVEL="INFO" \
-    MM_FILESETTINGS_DIRECTORY="/mattermost/data/" \
-    MM_FILESETTINGS_ENABLEPUBLICLINK=true \
-    MM_FILESETTINGS_PUBLICLINKSALT="A705AklYF8MFDOfcwh3I488G8vtLlVip" \
-    MM_EMAILSETTINGS_SENDEMAILNOTIFICATIONS=false \
-    MM_EMAILSETTINGS_FEEDBACKEMAIL="" \
-    MM_EMAILSETTINGS_SMTPSERVER="" \
-    MM_EMAILSETTINGS_SMTPPORT="" \
-    MM_EMAILSETTINGS_INVITESALT="bjlSR4QqkXFBr7TP4oDzlfZmcNuH9YoS" \
-    MM_RATELIMITSETTINGS_ENABLE=true \
-    MM_SQLSETTINGS_DRIVERNAME="postgres"
-
 # Install some needed packages
 RUN apt-get update \
     && apt-get -y install \
       curl \
+      jq \
       netcat \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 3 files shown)


jasonblais:
@pichouk Mattermost-Heroku uses these variables, and I haven't heard issues so far. 

Wondering if there is a bug in how Mattermost handles environment variables or if there's something that can be tweaked so they work in the Docker deployment too?

pichouk:
@jasonblais I look at the [mattermost/mattermost-heroku](https://github.com/mattermost/mattermost-heroku) repository and it doesn't seems to use Mattermost environment variables for configuration. I look quickly but it seems to use some environments variable (not Mattermost env var beginning with `MM_`) to substitute some parameters on a configuration template.

Actually, even if the environment variables was elegant, I think that the solution in this PR is ok : it will not be necessary to rebase a configuration file template everytime it changes. We just need to look at Changelog carefully in case Mattermost introduce a breaking change in `config.json` format (eg. change name of an existing key).

jasonblais:
Sorry, the official repository has actually moved here: https://github.com/mozilla/mattermost-heroku/. I'll submit an update to clarify that in the https://github.com/mattermost/mattermost-heroku repo.

The second option works too. Some are moving to use environment variables to pass configuration settings to MM instead of generating config.json because Mattermost also writes data there, and if a reconfigure occurs then these newer settings can be lost. I'm not sure how applicable that is to the Docker project.

pichouk:
So is this working with Mattermost-Heroku ?
Because when I use those env vars it really blocks any possibility of customizations from system console.

pichouk:
Just saw this on the Mattermost-Heroku readme :
```
Don't make configuration changes in the Mattermost admin console. Any configuration changes in the Mattermost admin console will be lost on dyno restart (which may be every 24 hours) and will not be distributed across multiple dynos
```
So it seems that they have the same behavior.

I think we are safe with this solution. The solution here will prevent to loose settings because :
- if using a volume for the `config` folder (which is the recommended way), configuration will persists and never get override
- if there is no configuration file the default config is used

jasonblais:
Sounds good 👍

pichouk:
I merge this PR since the image is currently broken and it should fix it. If someone came with an idea to improve this solution, we will open another PR.

PR #139, opened 2017-06-10
Do not use hardcoded value for salts 

pichouk:
In PR #31, @pierreozoux proposed an elegant way to not use hardcoded salts in app container. I propose to merge this to master.

(cc @MichaelEischer who also ask for this change)

### DIFF
--- a/app/docker-entry.sh
+++ b/app/docker-entry.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+generate_salt() {
+  cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 48 | head -n 1
+}
+
 DB_HOST=${DB_HOST:-db}
 DB_PORT_NUMBER=${DB_PORT_NUMBER:-5432}
 MM_USERNAME=${MM_USERNAME:-mmuser}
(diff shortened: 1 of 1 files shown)


app/docker-entry.sh
```diff
@@ -33,14 +37,27 @@ if [ "$1" = 'platform' ]; then
       jq '.LogSettings.ConsoleLevel = "INFO"' $MM_CONFIG > $MM_CONFIG.tmp && mv $MM_CONFIG.tmp $MM_CONFIG
       jq '.FileSettings.Directory = "/mattermost/data/"' $MM_CONFIG > $MM_CONFIG.tmp && mv $MM_CONFIG.tmp $MM_CONFIG
       jq '.FileSettings.EnablePublicLink = true' $MM_CONFIG > $MM_CONFIG.tmp && mv $MM_CONFIG.tmp $MM_CONFIG
-      jq '.FileSettings.PublicLinkSalt = "A705AklYF8MFDOfcwh3I488G8vtLlVip"' $MM_CONFIG > $MM_CONFIG.tmp && mv $MM_CONFIG.tmp $MM_CONFIG
+      jq '.FileSettings.PublicLinkSalt = "ChangePublicLinkSalt"' $MM_CONFIG > $MM_CONFIG.tmp && mv $MM_CONFIG.tmp $MM_CONFIG
```
xcompass: Can you just call `generate_salt` here? Then you don't need the logic from line 51 to 60.

pichouk: Good point ! I just modified this.

xcompass:
Thanks @pichouk. Great work!

Issue #142, opened 2017-06-16
getaddrinfo: Name or service not known for sloppy.io hosting


JannikZed:
Hello there,
I'm currently trying to get mattermost in docker running for a docker hosting-provider called sloppy.io. I managed it to connect app with db, web with app, but app is not answering requests at port 80... in the logs I see this entry every second: mattermost services app nc: getaddrinfo: Name or service not known
I don't know, which mattermost-service is causing this issue, but it would be nice, if someone could give me some "inspiration" what is causing this issue. 

Thank you

pichouk:
I don't know how [sloppy.io](https://sloppy.io) works, could you provide the configuration you pushed into it and, if possible, the entire log on Mattermost app ?

JannikZed:
Sloppy is Apache Mesos based. The difference from docker compose is, that you need more "fully qualified domain names" to address the containers. For example: At the web container I had to change proxy_pass http://app:80; to proxy_pass http://app.services.mattermost.jannikzinkl:80;
Here is the complete log I received: 

2017-06-16 15:31:28 mattermost services app No configuration file /mattermost/config/config.json
2017-06-16 15:31:28 mattermost services app Creating a new one
2017-06-16 15:31:28 mattermost services app Configure database connection...OK
2017-06-16 15:31:29 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:30 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:32 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:33 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:34 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:35 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:36 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:39 mattermost services app nc: getaddrinfo: Name or service not known
2017-06-16 15:31:40 mattermost services app nc: getaddrinfo: Name or service not known

That's all that is happening..
The data I insert is kind of the same than the docker-compose file out of this repo. This there a possibility to activate a debug-mode via environment variables? 
Do you have an idea, at which stage of the starting-process this strange "getaddrinfo: Name or service not known" is coming?

pichouk:
[...]

PR #144, opened 2017-06-19
Use links instead of depends_on

pichouk:
In `docker-compose.yml` file, I think it should be better to use `links` instead of `depends_on` by default. It will allow to change services names (like in https://github.com/mattermost/mattermost-docker/issues/142#issuecomment-309374431 ) without breaking containers which hardcode `app` and `db` hostnames.


pichouk:
@xcompass I'll wait for your approval on this, but I think it will not break anything.

PR #146, opened 2017-06-19
Merge app dockerfiles

pichouk:
Maybe it's just me, but I think maintaining 2 Dockerfiles for app is annoying and can be source of errors. Since the difference is just the download URL of Mattermost, I propose to use build argument to choose between between `team` and `entreprise` edition.
I didn't change the default behavior (which is `entreprise` edition), so this is not a breaking changes. The only change is that it need to comment out two lines on `docker-compose.yml` instead of only one to use `team` edition.

@jasonblais  @xcompass I'll appreciate reviews since there is, maybe, a good reason to keep 2 separate Dockerfiles. But I don't see it.


pichouk:
A dirty solution could be, when we want to trigger an auto-build (new version of Mattermost).
- modify app Dockerfile with `team` as default value, commit and tag as `NUMVERSION-team`
- modify app Dockerfile with `entreprise` as default value, commit and tag as `NUMVERSION-entreprise`

So Docker Hub images will have a `NUMVERSION-team` and `NUMVERSION-entreprise` tag and `latest` will alway be Entreprise edition.

xcompass:
That could work, but git history will be kind of confusing. Is it better than two dockerfiles?

pichouk:
Yes Git history will be a mess, you're right I don't think it is better than 2 Dockerfiles. 
Or maybe configure Travis to run a script only on tagged commit and build/push 2 versions (team and entreprise) of the image to Docker Hub. I'm not familiar with this (usually I use private docker registry and Gitlab CI) but I guess I could setup something like that.

pichouk:
Up @jasonblais , is there plan to build, for Docker Hub, team **and** enterprise edition ? Because if not, I'll merge this ;)

jasonblais:
Not sure, I don't think right now. If we did add it, would this cause issues in the future?

pichouk:
@jasonblais I don't know how Docker Hub auto-build works (I don't even know how it is configured for this repository). But @xcompass seems to say that it does not support build argument, so if we merge the 2 Dockerfiles (that's what this PR do) then you will only be able to build one version (Team or Entreprise) automatically to Docker Hub.

xcompass:
That's correct. If we ever want to build two versions on dockerhub, we have to have two dockerfiles. If you look at other official repos that has multiple variants, they have multiple dockerfiles. (https://github.com/docker-library/php/tree/master). To void multiple dockerhub repo for different versions, maybe we can apply for official repo on dockerhub, like others. (https://github.com/docker-library/official-images). Not sure if the team wants to go that direction. @jasonblais ?

jasonblais:
@pichouk @xcompass Excellent points.

@pichouk  On Monday, would you like to post in the [Developers channel](https://pre-release.mattermost.com/core/channels/peer-to-peer-help) and get the Mattermost developers thoughts? I'm thinking this question is better directed to our core engineers than myself.

If you don't hear any response, let me know and I'll bring it up to the team myself.

pichouk:
@jasonblais I'll do :)

pichouk:
@jasonblais Conclusion is that it seems people using Mattermost Enterprise Edition can use without Licence and it will be the exact same as Team Edition.

So I think we can have only Enterprise Edition on Docker Hub, so using Docker build argument is OK, so we can merge Dockerfiles.

jasonblais:
Yep, sounds good. As long as we offer a way to also install TE (which the above plan includes), we should be covered

jasonblais:
Thanks @pichouk :)

xcompass:
Thanks @pichouk

Issue #104, opened 2017-02-24
Support for docker Stack


thepill:
Would be great if you could supply Support/documentation for the docker stack mode.

Thanks!

xcompass:
Never played with docker stack. Would you be willing to submit a PR?

pichouk:
@pastukhov in #147 seems to say that using a docker-compose version 3 file should made I work with Docker Stack.
Since it seems that there is no breaking changes for us on docker-compose v3 ([see changes](https://docs.docker.com/compose/compose-file/compose-versioning/#upgrading)), I think we could move on this version.

pichouk:
Just opened a PR for this. But seems that dcoker-compose v3 required a "high" version of Docker Engine, not already supported on all distributions.

pichouk:
PR #176 introduced swarm mode support !

Issue #147, opened 2017-06-23
FR add ability to deploy with swarm


pastukhov:
Please upgrade docker-compose.yml to version 3.
It make able to run mattermost on swarm cluster with ``` docker stack deploy ```

pichouk:
Seems related with #104

pichouk:
@pastukhov Are you sure that supporting docker-compose version 3 will allow to use Docker stacks with Swarm ? 
https://github.com/mattermost/mattermost-docker/pull/148#issuecomment-326603278 seems to say it will not.

pdemagny:
@pichouk I have written & tested a swarm mode ready yml file, how can i share it with you ?

pichouk:
@pdemagny A Github gist can be a good way ;) Or a pastebin if you prefer.

pdemagny:
@pdemagny or a PR :)

pichouk:
@pdemagny Sure ! I'm interesting, but I understood [here](https://docs.docker.com/docker-cloud/apps/stacks/) that a Docker compose file should allow to use `docker stack`

pdemagny:
@pichouk that depends.
With docker-compose you mostly build and run on a single docker engine (usually your laptop) for development purposes.
The big difference with docker stack (swarm mode of the main docker engine binary) is that you can build images at stack creation but they would only live on a single node of your swarm cluster (the on where you run the docker stack deploy command), which is totally pointless.
Swarm is meant to run pre-built images of your micro-services spreaded across several docker engines (this is a better fit for production of course) as explained here: https://docs.docker.com/compose/swarm/#limitations

I've made a compose v3.3 file fully compatible with docker swarm mode for myself, utilizing your prod-app & prod-db docker hub images. Its no big deal but i'd be glad to contribute it on this project, if that's ok with you.
How should i proceed ? (i'm a sysadmin not a github guru, a little guidance would be nice).
Requirements & installation procedures differs, shall i update the README.md accordingly ?

pichouk:
@pdemagny Great !
If you have something working, you can fork this repository and add your modifications (docker-compose file and documentation) to your forked repository. Then, [submit a Pull Request](https://github.com/mattermost/mattermost-docker/pulls) and we will discuss and, I hope, merge your commits :)

pichouk:
Thanks to @pdemagny and PR #176, swarm mode is now supported !

PR #148, opened 2017-06-23
Move to docker-compose v3

pichouk:
Should allow support for docker stack. #104

We should have no issue with the `docker-compose.yml` file itself there is no changes on options we use.
The only concern is that [it seems to required](https://docs.docker.com/compose/compose-file/compose-versioning/#compatibility-matrix) at least Docker Engine `1.13.0` which is quite recent and, AFAIK,  not yet on Ubuntu official repositories.

Maybe we should add a second docker-compose file for v3 ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ The following instructions deploy Mattermost in a production configuration using
 * [docker]
 * [docker-compose]
 
+The `docker-compose.yml` file use Docker Compose `3.0+` format (which require Docker `1.13.0+`). If you want to use previous version of Docker or Docker Compose, you should modify the `version` field on the `docker-compose.yml` file.
+
 ### Choose Edition to Install
 
 If you want to install enterprise edition, you can skip this section.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "2"
+version: "3"
 
 services:
 


pichouk:
Even if it is easy to have a specific `docker-composeV3.yml` file, there is a real question here : what are our requirements ? Currently we only specify `docker` and `docker-compose` as requirements, but maybe we should specify versions number (like Mattermost which have OS/software/web client requirements).
@jasonblais @xcompass

pichouk:
I'll open an issue before merging because I think we should at least mention versions requirements on the README.

xcompass:
Sounds good. Thanks!

pdemagny:
Hi, correct me if i'm wrong, but supporting docker-compose v3 format does not mean you are docker stacks (swarm mode) compatible. For instance, links are not supported by swarm mode stacks, like a bunch of other suboptions (see https://docs.docker.com/compose/compose-file/#not-supported-for-docker-stack-deploy). That being said, a fully compatible docker-compose.yml v3 file for docker stacks would be nice !

pichouk:
@pdemagny I don't know, I just re-open #147 to discuss about that.

pichouk:
Closing since we don't need to support Compose file version 3 for now. Docker stacks support will come on another PR.

PR #152, opened 2017-07-10
Only configure database connection if needed

jnbt:
This allows MySQL users to predefine a custom `MM_SQLSETTINGS_DATASOURCE`, which needed for the database connection. 

In the official [docs](https://docs.mattermost.com/install/install-ubuntu-1604.html#installing-mattermost-server) the following connection „string“ is mentioned for MySQL users: 

```
"mmuser:<mmuser-password>@tcp(<host-name-or-IP>:3306)/mattermost?charset=utf8mb4,utf8&readTimeout=30s&writeTimeout=30s
```

Without this change the container forces the Mattermost server to use a Postgres connection, even if the config file defines MySQL.

### DIFF
--- a/app/docker-entry.sh
+++ b/app/docker-entry.sh
@@ -51,9 +51,14 @@ if [ "$1" = 'platform' ]; then
       echo "Using existing config file" $MM_CONFIG
     fi
 
-    echo -ne "Configure database connection..."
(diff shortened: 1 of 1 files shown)


jnbt:
@pichouk Fixed - I was too tired ☕️ 
By the way: Thanks for your work! ✨

PR #154, opened 2017-07-16
Improve README

pichouk:
I propose a re-organisation of the README file. I thought it is better like this, open to any suggestion/disagreement ;)

The Web container doc could still be improve, but I think its entrypoint could be too, so I leave this for another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ If you use a Mattermost configuration file on a different location than the defa
 * `MM_CONFIG`: configuration file location inside the container.
 
 If you choose to use MySQL instead of PostgreSQL, you should set a different datasource :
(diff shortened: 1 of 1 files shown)


jasonblais:
@pichouk 

> I thought it is better like this, open to any suggestion/disagreement ;)

Would you have a few sentences on how this improves the README? Does it make it easier to install the Docker for instance?

pichouk:
@jasonblais I mainly reorganize sections of the README, without removing important pieces of information.
I thought it is more interesting to organize the documentation by using separate sections explaining the 3 different components (db, app and web front containers).

Inconvenient :
- You have to read a more lines of doc (not that much) to "quickstart".

Advantages :
- You directly understand how different components works together.
- I think it can avoid using default (and unsecure) database password by putting the information about this customization at the beginning.
- I think it is more clear (but this is really subjective)

pichouk:
Any objections ? I'll merge otherwise ;)

xcompass:
LGTM 👍

Issue #156, opened 2017-07-19
ERR_TOO_MANY_REDIRECTS with "Forward port 80 to 443" option set to True


jminardi:
Steps to reproduce:

- Spin up a brand new Ubuntu 16.04 EC2 instance
- Install docker and docker-compose
- `git clone https://github.com/mattermost/mattermost-docker.git`
- generate SSL certs with certbot and copy them to `mattermost-docker/volumes/web/cert/`
- `docker-compose up -d`
- `docker-compose start`
- navigate to https://my.domain (note that HTTPS is working at this point)
- add https://my.domain as the site url as requested by the banner
- Find "Forward port 80 to 443" option and set it to True
- restart mattermost (`docker-compose down && docker-compose up -d`)
- Navigate to https://my.domain
- See the `ERR_TOO_MANY_REDIRECTS` in chrome and a similar error in safari

Any help is appreciated.

NOTE: I tried clearing my cookies as suggested by chrome, but that didn't help

pichouk:
Your `app` container is listening to port `80`.
Your `web` container is listening to port `443` since you enable TLS. It works like a reverse proxy and redirect all requests to your `app` container, on port `80`.

If you enable the "Forward port 80 to 443" on your `app` container, then **all** your requests will reach your `web` container on `443` and be redirect to `app` on `80`. So the `app` container redirect you to port `443` so your client made a new request on `web` (port `443`) which redirect to `app` (port `80`) and so on...

I think you should not enable this option if you use the `web` container or any kind of reverse proxy to reach Mattermost.

jminardi:
Thats for the quick response @pichouk, I think I follow.

It seems like I should instead be setting up the forward on the `web` container. Do you know the best way to do that? I assume there is some sort of configuration file for that I could edit.

I found this file: https://github.com/mattermost/mattermost-docker/blob/master/web/mattermost-ssl

Do you know how I use that, and if I can modify it to redirect?

pichouk:
Try to add this at the beginning of the file you mention :
```
server {
	listen 80 default_server;
	listen [::]:80 default_server;
	server_name _;
	return 301 https://$host$request_uri;
}
```

Then build again `docker-compose build` and restart containers. Could be interesting to merge this on `master` if it works.

jminardi:
That worked!

I think it would make sense to merge something like this into the repo.

pichouk:
[...]

PR #157, opened 2017-07-19
Add 301 redirect from port 80 to port 443

jminardi:
As suggested by @pichouk here: https://github.com/mattermost/mattermost-docker/issues/156#issuecomment-316475874


pichouk:
LGTM 👍 
I'll wait for a @xcompass approval since it change a default behavior.

PR #181, opened 2017-10-14
Allow to customize application host/port on Web container

pichouk:
On #142 issue @JannikZed explained that, because he was using Sloppy.io provider, he had to modify Web container to use a different App hostname.
Because it cost nothing to add this feature, I just add few lines and Web image to allow user to choose (with environment variables) the hostname and port of the app container. Default values stay `app` and `80`, so not a breaking change.

As usual, I appreciate review @xcompass ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,9 @@ If you choose to use MySQL instead of PostgreSQL, you should set a different dat
 Don't forget to replace all entries (beginning by `MM_` and `DB_`) in `MM_SQLSETTINGS_DATASOURCE` with the real variables values.
 
 ### Web server container
-This image is optional, you should not use it you have your own reverse-proxy. It is a simple front Web server for the Mattermost app container.
-* `MATTERMOST_ENABLE_SSL`: whether to enable SSL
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -74,9 +74,9 @@ If you choose to use MySQL instead of PostgreSQL, you should set a different dat
 Don't forget to replace all entries (beginning by `MM_` and `DB_`) in `MM_SQLSETTINGS_DATASOURCE` with the real variables values.
 
 ### Web server container
-This image is optional, you should not use it you have your own reverse-proxy. It is a simple front Web server for the Mattermost app container.
-* `MATTERMOST_ENABLE_SSL`: whether to enable SSL
-* `PLATFORM_PORT_80_TCP_PORT`: port that Mattermost image is listening on
+This image is optional, you should **not** use it you have your own reverse-proxy. It is a simple front Web server for the Mattermost app container. If you use the provided `docker-compose.yml` file, you don't have to configure anything. But if your application container is reachable on custom host and/or port (eg. if you use a container provider), you should add those two environment variables :
```
xcompass: Add "when" `after you should **not** use it`

PR #185, opened 2017-10-19
Remove deprecated links in docker-compose file

sebgl:
Using links is considered to be a deprecated feature of Docker,
in favor of user-defined networks (see
https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/).

Since the docker-compose file is in Version 2, a network is
automatically created for the 3 containers, making them accessible to each others with their name directly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,8 +37,6 @@ services:
       - MM_DBNAME=mattermost
       # in case your config is not in default location
       #- MM_CONFIG=/mattermost/config/config.json
-    links:
-      - db:db
 
   web:
     build: web
@@ -53,5 +51,3 @@ services:
     # Uncomment for SSL
     # environment:
     #  - MATTERMOST_ENABLE_SSL=true
-    links:
-      - app:app


xcompass:
@pichouk I remember we left it there because someone was having conflict with default name "db". So that they can change to something else. You made this PR a while ago: https://github.com/mattermost/mattermost-docker/pull/144 switch from depends_on to links. Considering docker is moving away from docker-compose towards docker stack, we pulled in your PR. Is it good idea to remove links?

pichouk:
Since `links` is the old manner, it is better to remove this.

When I submitted PR #144 some hostnames where hardcoded on `app` and `web` containers. So changing the name of `app` or `db` container would have broke things. Now we can change those values with environment variables (`DB_HOST` in `app` container for example) so I guess `links` entries are not necessary anymore.

xcompass:
Good point. But I'm not sure if docker-compose link is the same as docker link. The reference is for docker link, which will be deprecated. I can't seem to find any deprecation info for docker-compose link. 

I have no problem removing it as I don't generally include it in my docker-compose files either. Just want to clarify things to avoid future confusion.

pichouk:
That's a good point. I imagine that Docker Compose `links` is based on Docker `link` flag. I think it is, because when I run some containers with Docker Compose there is a `Links` entry in `NetworkSettings` when I look at `docker inspect`. Seems to be the same for Docker `--link` flag.
I didn't find documentation about this, but maybe looking at the [source code](https://github.com/docker/compose) will help to determine if those options are the same.

PR #187, opened 2017-10-25
Add version 4.0.5

xcompass:
We need to upgrade to 4.0.5 for now. Here is the change log: https://docs.mattermost.com/administration/changelog.html#release-v4-0-5

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:14.04
 
 # Some ENV variables
 ENV PATH="/mattermost/bin:${PATH}"
-ENV MM_VERSION=4.3.1
+ENV MM_VERSION=4.0.5
 
 # Build argument to set Mattermost edition
 ARG edition=entreprise


xcompass:
Do we want all the versions that upstream support currently to be tagged?

pichouk:
IMHO we need to stick to latest stable release (here `4.3.1`) on `master` branch.
I don't see any problem if we maintain branches for "old" versions (here `4.0.x`n `4.1.x` and `4.2.x`) as long as it doesn't impact `master` branch. But is that really helpful for people ? If we maintain branches just with one little change (which is version number on app Dockerfiles) I don't see the point. If we manage to deploy those images to Docker Hub, then it could be a good idea. Does tags on "non-master" branches trigger Docker Hub build (cc @jasonblais) ?

To sum up, I agree if :

xcompass:
It won't affect master as long as we do PRs to the version branch (like in this PR), e.g. 4.0.x, 4.1.x, 4.2.x. We can then tag those versions and it will be built by dockerhub as all tags will be built. People who needs older version and don't want to build themselves can pull from dockerhub directly.

PR #176, opened 2017-09-23
add docker swarm mode support

pdemagny:
Signed-off-by: Pierre DEMAGNY <pdemagny@printoclock.com>

Issue #151, opened 2017-07-06
Docker requirement for this repository


pichouk:
Before to merge PR #148, I would like to open a discussion about the Docker requirement of this repository. Mattermost have OS/software/web client requirements, and I guess we should add a Docker requirement on our repository too. It should be better if we are clear on what versions we support, what version we will no longer support, and will allow to identify and announce breaking changes (eg. using a new functionality of `docker-compose`).

To begin, I think we can add versions requirement for `docker-engine` and `docker-compose` on the README. I propose :
- `1.13.0+` for Docker
- `3.0+` for Docker Compose

pichouk:
Any opinion on this @xcompass or @jasonblais ?

jasonblais:
@pichouk I agree we should add version requirements in README, perhaps even to our install guides
 - https://docs.mattermost.com/install/prod-docker.html
 - https://docs.mattermost.com/install/docker-local-machine.html

As for the actual Docker version numbers, I don't have a strong opinion.

Would you have thoughts on how many might be affected if we restrict to 1.13.0+ for Docker and 3.0+ for Docker Compose

pichouk:
Docker `1.13.0+` allow to use docker-compose `3.0+`, which allow to support the "future" of docker-compose : docker stack. And we have some [demands about this](https://github.com/mattermost/mattermost-docker/issues/104).
**BUT** I know that some people use the Ubuntu packaged Docker binary instead of the official (called `docker.io`). I don't know how many people it affect, but this package is, AFAIK, on `1.12.x`. So it could break things for those people.

From here, we have 3 solutions :
- we support docker-compose `3.0+` and we consider people should have an almost up-to-date Docker version
-  we do not support docker-compose `3.0+` so we could announce "only" Docker `1.12+` as requirement
- we announce Docker `1.12+` but we propose both `2.0` and `3.0` docker-compose files
 
According to me, maintaining 2 docker-compose files is dirty since, for now, the only difference is the first line, specifying the version number to use. But I also understand that some people may use sensitive systems and prefer to wait few months before updating Docker, and `1.13.0` is only 7 months old.

If I was the only one to decide, I will use the `1.13.0+` and `3.0+` requirements. I guess this is not a really strong and complicated to achieve requirements, but maybe some people will need an update (I have no metrics to say how much). But I can understand if Mattermost want to be more permissive about requirements.

jasonblais:
[...]

PR #195, opened 2017-11-11
Updated docker-compose to 3.0, use of .env file, and use of a custom network

joserprieto:
Hi;

This may be useful for your;

- I updated the docker-compose.yml file, to docker-compose 3.0 specs.
- Avoid the use of "enviroment" key, and isolate the enviroment vars in a separated file (I versioned a .env.example only).
- Use of a custom network, for better management in the host running docker.

If you want, i can edit the readme and document the use and changes, and why is better this approach.

Regards.


joserprieto:
Tested in:

Linux Mint 18.2, and last versions of Docker and docker-compose.

Debian 9 and last versions of Docker and docker-compose.

pichouk:
Thanks @joserprieto !
I'm quite busy for now but I promise I'll take a look :)

docker-compose.yml
```diff
@@ -1,53 +1,68 @@
...
+    container_name: "mattermost_app"
     build:
       context: app
       # comment out 2 following lines for team edition
-      # args:
-      #   - edition=team
+      args:
+        - edition=team
     restart: unless-stopped
+    links:
```
pichouk: We removed `links` in PR https://github.com/mattermost/mattermost-docker/pull/185 because it is deprecated. Should remove it I guess.

pichouk:
So we're back on the same question again @jasonblais and @xcompass : should we move to Docker Compose 3 ?

Maybe it is a good idea. Actually, the `docker-compose.yml` file provided in this repository is just a cool tool we provide to help people deploying with Docker. In fact, it is quite easy to build Docker images (using Dockerfiles) and deploy without Docker Compose (or with an old version). I did it for my production server because I have some custom needs, and this is not something really hard **if you have some basics with Docker**.
My point is : the `docker-compose.yml` file on this repository is not that critical. In few minutes you can easily create a new one, matching your needs. So changing version of Docker Compose is okay as soon as we provide correct documentation on how to build the Docker images.

jasonblais:
@pichouk I forget the previous discussion, but what were the potential side effects of moving to Docker Compose v3? (Link to a previous post is okay, I wasn't able to find it)

And conversely, what is the benefit of moving to v3? At first it was to support swarm cluster I think, but that is resolved with https://github.com/mattermost/mattermost-docker/pull/176

docker-compose.yml
```diff
@@ -1,53 +1,68 @@
-version: "2"
+version: '3'
```
xcompass: Which docker-compose directive requires v3 format? If we upgrade the compose files to v3, it will break backward compatibility. Not sure if it is good time to upgrade to v3 and if we decide to go with v3, we need to list it as a breaking change.

pichouk:
@jasonblais We add some discussions about this subject on #151 and [this forum post](https://forum.mattermost.org/t/mattermost-with-docker-discussion-about-requirements/3518) (and a little on #104 and #147).

Yes the main benefit to moving to v3 is the swarm support, but it was solved. So, the "only" benefit now is to be up to date on our default configuration to deploy with Docker. I have no particular interest to moving on v3. It just not the first time someone submit the idea ;)

jasonblais:
Thanks! I'm okay switching to v3, don't see major issues and seems to be the preference of the community @pichouk

jasonblais:
As long as we update the documentation accordingly, which you've taken care of here :) 

Can you also help take a look at 
 - https://docs.mattermost.com/install/prod-docker.html and 
 - https://docs.mattermost.com/install/docker-local-machine.html

to see if anything needs to be updated?

pichouk:
@jasonblais Great :)
We will work on this PR with @joserprieto since there is some things to correct/change, but when it will be ready I'll take particular attention to Mattermost documentation and will update it :)

docker-compose.yml
joserprieto: The links is because you can assign one of more aliases for the different hosts; i think that, for the moment, docker-compose will mantain retrocompatibility

docker-compose.yml
```diff
@@ -1,53 +1,68 @@
...
+  mattermost_app:
+    hostname: "app.mattermost.local"
+    container_name: "mattermost_app"
     build:
       context: app
       # comment out 2 following lines for team edition
-      # args:
-      #   - edition=team
+      args:
+        - edition=team
```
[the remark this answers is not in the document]
joserprieto: Ok.
For, then, the docker-compose.yml don't be versioned.

joserprieto:
I read all the discussions, and i think that here are a debate of use version 2 or version 3 of docker-compose files.

The solution can be easy; put docker-compose.v2.yml.sample and docker-compose.v3.yml.sample, and the correspondent instructions on the README.

I make a PR right now with this approach in mind.

docker-compose.yml
```diff
@@ -1,53 +1,68 @@
...
-      - POSTGRES_PASSWORD=mmuser_password
-      - POSTGRES_DB=mattermost
-    # uncomment the following to enable backup
-    #  - AWS_ACCESS_KEY_ID=XXXX
-    #  - AWS_SECRET_ACCESS_KEY=XXXX
-    #  - WALE_S3_PREFIX=s3://BUCKET_NAME/PATH
-    #  - AWS_REGION=us-east-1
+    env_file:
+      - .env
+    networks:
```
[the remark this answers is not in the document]
joserprieto: For many reasons, you can want to have under control the IPs and the ports exposed outside, and so on.

joserprieto:
Changes done!

Please, review, @pichouk @xcompass 

Regards.

joserprieto:
@pichouk about the main documentation, that remarks @jasonblais, i think that:

https://docs.mattermost.com/install/prod-docker.html

> Deploy the Mattermost Production Docker setup by running:
> 
> git clone https://github.com/mattermost/mattermost-docker.git
> cd mattermost-docker
> docker-compose build
> docker-compose up -d

Can be:

> Deploy the Mattermost Production Docker setup by running:
> 
> git clone https://github.com/mattermost/mattermost-docker.git
> cd mattermost-docker
> open the README and follow instructions to select docker-compose version and for the enviroment vars.
> docker-compose build
> docker-compose up -d

The other link:

https://docs.mattermost.com/install/docker-local-machine.html

Is for use on a dev environment, not for make all the stack and so on.

Regards!

xcompass:
@joserprieto Thanks. I would prefer to leave a docker-compose yaml file as default one (maybe v3 one). And let people to copy v2 one or use `-f` option if they need v2 compose file.

joserprieto:
@xcompass so, change this?

I not like the versioning of "production" files, because if you want to change this, you will not can update the repo (imagine, you use this repo as a submodule...).

Opinions, @pichouk @jasonblais ?

pichouk:
As I suggest on https://github.com/mattermost/mattermost-docker/issues/203#issuecomment-348707187 I think we can have a "classic/old-school" `docker-compose.yml` file (for retrocompatibility) and add a new `docker-stack.yml` file (based on existing [`/contrib/swarm/docker-stack.yml`](https://github.com/mattermost/mattermost-docker/blob/master/contrib/swarm/docker-stack.yml)) for Swarm/Docker compose v3 features.

joserprieto:
There is a problem with maintain the docker-compose.yml versioned.

If i have one project, or, i want to deploy this project to a server, and i want to update, i can't change nothing of the repo; so, if i need adjust the docker-compose.yml file (for my needs, or comment the file, or something...), i can't.

I think that isn't good for the use of the repo.

So, we can have:

- `docker-compose-v3.yml` (or `docker-stack-yml`)
- `docker-compose-v2.yml`

And in the `REAME.md`, add instructions, like:

> To have v3 of docker-compose:
> 
> ln -s docker-compose-v3.yml docker-compose.yml
> 
> Or, if you want to customize your docker-compose file:
> 
> cp docker-compose-v3.yml docker-compose.yml

(Something like that).

If we provided "configuration files" (the docker-compose file defines and configure the stack that uses an application), the process of update isn't possible (as i think).

So, i think that a better approach, is 

1. have two files like this:

- `docker-compose-v3.yml` (or `docker-stack-yml`)
- `docker-compose-v2.yml`

2. add the docker-compose.yml to .gitignore

What do you think?

Regards.

(i think that you have to take one decision, @pichouk, @xcompass, and @jasonblais)

pichouk:
I agree with @joserprieto now. We will never achieve to have a "perfect `docker-compose.yml` that match everyone needs". Maybe we should provide multiple example files and let users choose what they want to use.
Also it will force users to understand what they are doing, and avoid to just run `git clone && docker-compose up -d`, that can lead to situations where a user doesn't understand how to upgrade/where data is stored, etc.

It will not be a breaking change. Current users will still have their `docker-compose.yml` file so it will not break their setup.

xcompass:
@pichouk when you remove the docker-compose.yml from this repo, wouldn't it also remove from the local repo when one pulls the change (assume no change in the file)?

pichouk:
@xcompass Hum you're right. I just tested it and if we want to remove the `docker-compose.yml` file and add it to `.gitignore`, it will remove the file on local repositories when pulling **only if you never change the file**.
If you modify this file (to change default password for example) you will not be able to pull, so it will not remove it. So everybody have to backup the file, pull the change, and restore it.

joserprieto:
I'm agree with @pichouk.
And, remember, this docker / stack is for production purpose; so, the normal behaviour is to change default passwords on the docker-compose.yml (passwords that will be isolated, on a .env file, like in my PR).

I think that a votation may resolve this question; logically, i can't vote, but the mattermost team's votes can resolve all of this.

Regards.

pichouk:
Good for me. If @jasonblais and @xcompass are ok, then you can update your PR to match this new organization.
We will take care that it didn't broke current users installations and that documentation is correctly updated.

joserprieto:
About of to not broke current users installations; i can see two scenarios:

User changed the docker-compose.yml file (the default behaviour, i hope...); so, this user can't make a git pull and not loses his configuration (on his docker-compose.yml file); when he makes a git pull, it has to force it, and discard changes on his docker-compose.yml file (so, this new approach is good for him).
User not changed the docker-compose.yml file; yes, in this case, if we put the docker-compose.yml on the .gitignore file, they docker-compose.yml file will "disappears" whith git pull.. But i think that is a very helpful behaviour, because forces the user to change default users and passwords (that are in the docker-compose.yml file already).

A correct approach (imho), it's to isolate the vars outside the docker-compose.yml file (on a .env file gitignored too), and isolate the production ready docker-compose.yml, that defines the stack (so, it's a configuration file, in a certain way..)

I think that is this approach is reflected on the documentation, the user will not have trouble with the change.

Regards.

pdemagny:
Any news on a merge for this PR ? Once its merged i'll start to work on #203 :)

pichouk:
@joserprieto can submit changes describe in [this comment](https://github.com/mattermost/mattermost-docker/pull/195#issuecomment-349274915) and, as soon as @xcompass and/or @jasonblais approve this changes, we can merge :)

joserprieto:
The use of docker volumes, or Traefick, i think that is a personal election.

For example; on the current stack definition, we have:

```yaml
    volumes:
      - ./volumes/app/mattermost/config:/mattermost/config:rw
      - ./volumes/app/mattermost/data:/mattermost/data:rw
      - ./volumes/app/mattermost/logs:/mattermost/logs:rw

    volumes:
      # This directory must have cert files if you want to enable SSL
      - ./volumes/web/cert:/cert:ro
```

And yes, we can use something like that:


```yaml
version: "3"
services:
(...)
    volumes:
      - mttm-config:/mattermost/config:rw
      - mttm-data:/mattermost/data:rw
      - mttm-logs:/mattermost/logs:rw
      - etc-localtime:/etc/localtime:ro
(...)

volumes:
  mttm-config:
  mttm-data:
  mttm-logs:
  web-cert:
  etc-localtime:
```

But it's the same of the discussion on other threads; is better to use volumes? is better to use custom networks? is better to use compose files v2 or v3?

The answer is easy; depends of the use case.

With traefick, is the same.

I think that put sample docker-compose files, is the best choice; and, with this, any user can take is own decissions, taking his/her use case.

joserprieto:
@pichouk , if i'm not wrong, i make the changes.

I resume changes made:

- Created two separated files; docker-compose-v3.yml.sample and docker-compose-v2.yml-sample.
- Updated the v2 to get the advantange of isolated environment vars in a .env file.
- Updated the README with the instructions to use the docker-compose that you choose, and the .env file that you need.
- gitignored files .env and docker-compose.yml

I think that is all done... But, if any change is needed, please, tell me.

joserprieto:
Another changes that can will add, is the explanation of how isolated the .env file and docker-compose.yml file.

One option (as valid as another), is take advantage of git hooks. And do something like:

- Put the .env and the docker-compose.yml files outside the repo directory.
- Make symlinks to this files, on the repo dir.
- Before a pull/push, remove this symlinks.
- After the pull/push, make the symlinks again.

Another form, is make a simple bash script outside the repo dir, that do something like "setup.sh --enable" and "setup.sh --disable", for put symlinks / delete symlinks, before/after pull/push.

Maybe useful to explain this on the README. Or, it can be useful on the docs. As you wish :wink:

pdemagny:
Well, for "simple" compose files (debate v2/v3), i think that going with the flow is important here, v2 is getting really old and should be present only for backward compatible reasons.

Regarding the "Volumes" debate, a quick look at docker doc is enough for me:
> Volumes are the preferred mechanism for persisting data generated by and used by Docker containers. While bind mounts are dependent on the directory structure of the host machine, volumes are completely managed by Docker. Volumes have several advantages over bind mounts ...


I read a lot of people being reluctant to move to docker volumes because they think its easier to backup, why would it be ? Docker volumes are stored in a predictible directory on the host ... (/var/lib/docker/volumes) and are named like you named them in the compose file.
They offer better integration with docker, better isolation & avoid dirty permissions hacks.
And btw using local mounts tied to one host makes no sense in a multi-node swarm where you want your storage to be distributed & redundant over multiple nodes, or at least portable.

Same goes for custom networks, they offer better isolation for services, while still allowing them to talk to each other if the user wants to.

As for the env file, to the best of my knowledge it is simply not functional in docker stacks for docker swarm mode, people have to use docker secrets & environment.

And for Traefik, yes it is a personal choice, but it offers this choice to the users of this repo, and doesn't prevent from adding more choices for them in the future by adding more compose/stack file with support for more load-balancers ...
Having a single node swarm is almost pointless, people running swarms have these kind of services, we simply provide them with a good example here.
Traefik is a container ready solution, unlike things such as nginx-proxy, and offers a quick and nice way of handling https by automating creation & renewal of ssl certificates.

For all of these reasons, i completely back the solution with multiple yaml "sample" files in the root of the repo, with something like:
- docker-compose.yml (v3, volumes)
- docker-compose-v2.yml (v2, mounts, for backward compatibility)
- docker-stack.yml (v3 with volumes, configs, secrets, environment, network)
- docker-stack-with_traefik.yml (same but with traefik & letsencrypt)

I'm willing to do this, or part of it, depending of what you decide.

.env.sample
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# See https://docs.docker.com/compose/environment-variables/#the-env-file
+
+# PostgreSQL Enviroment Variables:
```
pichouk: Typo `Enviroment` => `environment`

.env.sample
```diff
@@ -0,0 +1,29 @@
...
+POSTGRES_USER=mmuser
+POSTGRES_PASSWORD=mmuser_password
+POSTGRES_DB=mattermost
+#PGDATA="/path/to/your/postgresql/data/files | default: /var/lib/postgresql/data/pgdata optional"
+#POSTGRES_DB="name_of_your_postgresql_db | default: postgres_user value | optional
+# uncomment the following to enable backupe
```
pichouk: backup*

docker-compose.v2.yml.sample
```diff
@@ -50,4 +41,4 @@ services:
       - /etc/localtime:/etc/localtime:ro
     # Uncomment for SSL
     # environment:
-    #  - MATTERMOST_ENABLE_SSL=true
+    #  - MATTERMOST_ENABLE_SSL=true
```
pichouk: Please keep the newline :)

.gitignore
```diff
@@ -1,8 +1,15 @@
...
 .elasticbeanstalk/*
 !.elasticbeanstalk/*.cfg.yml
 !.elasticbeanstalk/*.global.yml
 
 # Generated data
 volumes
+
+# Specific files:
+
+docker-compose.yml
```
pichouk: Please keep a newline :)

README.md
```diff
@@ -30,6 +40,45 @@ args:
...
+
+We give you two docker-compose sample files, one supports version 3.0 of docker-compose, and the other one supports version 2.0 of docker-compose (you can see more of docker-compose file versions [here](https://docs.docker.com/compose/compose-file/), [here](https://docs.docker.com/compose/compose-file/compose-file-v2/), and [here](https://docs.docker.com/compose/compose-file/compose-versioning/)).
+
+You have to get in mind the requirements for v2 and for v3, and the features that has the v3 over v2 (swarm support, and others); you can see more on [upgrading version 2.x to 3.x ](https://docs.docker.com/compose/compose-file/compose-versioning/#upgrading), and discussions about it:
+
+* https://github.com/mattermost/mattermost-docker/issues/104
+* https://forum.mattermost.org/t/mattermost-with-docker-discussion-about-requirements/3518
+* https://github.com/mattermost/mattermost-docker/issues/151
+
+You only have to do is:
```
pichouk: `is` -> `this`

README.md
```diff
@@ -83,6 +132,14 @@ Put your SSL certificate as `./volumes/web/cert/cert.pem` and the private key th
 no password as `./volumes/web/cert/key-no-password.pem`. If you don't have
 them you may generate a self-signed SSL certificate.
 
+### Building images
+
+Before start the containers, we must to build the images:
```
pichouk: `we must to build ` -> `you must build`

pichouk:
@joserprieto Great :)
I made a review, I'll like @xcompass to review too.
Also, it is important to update documentation on [this page](https://docs.mattermost.com/install/prod-docker.html). So if you can submit a PR on [this file](https://github.com/mattermost/docs/blob/master/source/install/prod-docker.rst) too I'll appreciate :)

To add new features (as @pdemagny suggest) I prefer to wait for another PR, but it is a good idea :)

docker-compose.v2.yml.sample
joserprieto: done

README.md
joserprieto: done

README.md
joserprieto: done

.gitignore
joserprieto: done

.env.sample
joserprieto: done

.env.sample
joserprieto: done

joserprieto:
@pichouk i think that i made all fixes.
Review and if you see more errors, please, tell me.

About the documentation, i can propose here the new doc page, and if you like it, i wll make the PR.

Do you agree in proceed on this way?

Regards.

pichouk:
LGTM, thanks a lot :)
@joserprieto I'm okay to do it this way !

joserprieto:
But, this PR is still open?
Why?

pichouk:
@joserprieto You proposed to submit a PR to Mattermost documentation before merging this one.

joserprieto:
Ah, ok.
I'll try to do it this weekend

hanzei:
Hey @pichouk,

What is the status of this PR? Can it be merged after it has been synced with `master`?

pichouk:
We are [dropping the support of the `docker-compose.yml`](https://github.com/mattermost/mattermost-docker/pull/366) file so I guess we can close this PR.

PR #197, opened 2017-11-16
Upgrade to 4.4.1

andruwa13:


pichouk:
Thanks @andruwa13 ! You're way faster than me ^^

PR #199, opened 2017-11-18
add security config to nginx

andruwa13:


web/security.conf
```diff
@@ -0,0 +1,33 @@
...
+# 'soon' on Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=471020
+add_header X-Content-Type-Options nosniff;
+
+# This header enables the Cross-site scripting (XSS) filter built into most recent web browsers.
+# It's usually enabled by default anyway, so the role of this header is to re-enable the filter for 
+# this particular website if it was disabled by the user.
+# https://www.owasp.org/index.php/List_of_useful_HTTP_headers
+add_header X-XSS-Protection "1; mode=block";
+
+# with Content Security Policy (CSP) enabled(and a browser that supports it(http://caniuse.com/#feat=contentsecuritypolicy),
```
pichouk: You can remove everything from here since it is commented :)

pichouk:
Good idea to add some security basics to our Nginx configuration !
Except for some commented line removal, it looks good to me :)

web/security.conf
andruwa13: Delete **add_header X-XSS-Protection "1; mode=block";** ?

pichouk:
Arf sorry I think I explained badly my point.

I think you should keep all commented lines since they explained each options. But at the end of the file, some comments were useless since header `Content-Security-Policy` is not used. So I mean something like this :
```
# don't send the nginx version number in error pages and Server header
server_tokens off;

# config to don't allow the browser to render the page inside an frame or iframe
# and avoid clickjacking http://en.wikipedia.org/wiki/Clickjacking
# if you need to allow [i]frames, you can use SAMEORIGIN or even set an uri with ALLOW-FROM uri
# https://developer.mozilla.org/en-US/docs/HTTP/X-Frame-Options
add_header X-Frame-Options SAMEORIGIN;

# when serving user-supplied content, include a X-Content-Type-Options: nosniff header along with the Content-Type: header,
# to disable content-type sniffing on some browsers.
# https://www.owasp.org/index.php/List_of_useful_HTTP_headers
# currently supported in IE > 8 http://blogs.msdn.com/b/ie/archive/2008/09/02/ie8-security-part-vi-beta-2-update.aspx
# http://msdn.microsoft.com/en-us/library/ie/gg622941(v=vs.85).aspx
# 'soon' on Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=471020
add_header X-Content-Type-Options nosniff;

# This header enables the Cross-site scripting (XSS) filter built into most recent web browsers.
# It's usually enabled by default anyway, so the role of this header is to re-enable the filter for 
# this particular website if it was disabled by the user.
# https://www.owasp.org/index.php/List_of_useful_HTTP_headers
add_header X-XSS-Protection "1; mode=block";
```

Also I'll apreciate if you can squash all your commits in just one. Not really important if you don't I'll do it at merge time ;)

Thanks :)

web/security.conf
```diff
@@ -1,33 +1,4 @@
...
-# http://msdn.microsoft.com/en-us/library/ie/gg622941(v=vs.85).aspx
-# 'soon' on Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=471020
 add_header X-Content-Type-Options nosniff;
-
-# This header enables the Cross-site scripting (XSS) filter built into most recent web browsers.
-# It's usually enabled by default anyway, so the role of this header is to re-enable the filter for 
-# this particular website if it was disabled by the user.
-# https://www.owasp.org/index.php/List_of_useful_HTTP_headers
 add_header X-XSS-Protection "1; mode=block";
-
```
pichouk: Remove only the lines below. Not other commented before :)

andruwa13:
fix

pichouk:
Great ! Thanks again :)

PR #201, opened 2017-11-19
Images Ubunutu 17.04, optimization and reduce image size

andruwa13:
17.04 more secure image than the outdated 14.04, the image size went from **108MB** to **82MB** (New Version)

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:17.04
 
 # Some ENV variables
 ENV PATH="/mattermost/bin:${PATH}"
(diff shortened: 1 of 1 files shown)


app/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:17.04
```
pichouk: Good idea to upgrade, but I'll prefer to use LTS version, so `16.04` instead of `17.04`.

app/Dockerfile
```diff
@@ -8,11 +8,12 @@ ENV MM_VERSION=4.4.1
 ARG edition=entreprise
 
 # Install some needed packages
-RUN apt-get update \
-    && apt-get -y install \
+RUN apt-get update && \
```
pichouk: Please keep the `&&` part on the newline to be consistent with other Dockerfiles :)

app/Dockerfile
```diff
@@ -8,11 +8,12 @@ ENV MM_VERSION=4.4.1
...
+    apt-get -y --no-install-recommends --allow-unauthenticated install \
```
pichouk: Why using `--allow-unauthenticated` ? Seems that it bypasses signature checks.

pichouk:
Good idea to upgrade Ubuntu (in fact my secret goal is to use Alpine based image, but we need a lot more tests).
I posted a review on some points, i'm open to discussion :)

Also, I think you made a mistake because there is 2 commits from your other PR in this PR. You should remove them ;)

andruwa13:
not work alpine images :( because he made an image on 17.04
rancher 2017-11-20 01-49-48

pichouk:
Yeah, I know that Alpine is not that easy. Let's go for Ubuntu `16.04` (LTS) for now.

PR #204, opened 2017-11-20
Images Ubunutu 16.04, optimization and reduce image size

andruwa13:


pichouk:
Great, thank you ! :)

PR #205, opened 2017-11-21
Images Alpine, optimization

andruwa13:


pichouk:
Thanks for your help, but I'll prefer if we split change to `db` and `app` images.
I think it is ok to use Alpine for our database image, just need to test it. But `app` image is way more sensitive since Alpine is not officially supported by Mattermost, so we will need to discuss this also with Mattermost staff.
So it is better to have a PR for database, which can be quickly merge **if tests worked**, and another one for `app` which need more discussions :)

andruwa13:
@pichouk app work, I tested everything on my test server

PR #208, opened 2017-11-21
Images App Alpine, optimization

andruwa13:


app/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM alpine
```
pichouk: We need to stick on a specific Alpine version. Let's go for `alpine:3.6`, which seems to be the latest.

app/Dockerfile
```diff
@@ -7,15 +7,20 @@ ENV MM_VERSION=4.4.1
...
-RUN apt-get update \
-    && apt-get -y --no-install-recommends install \
-      curl \
-      jq \
-      netcat \
-      ca-certificates \
-    && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin
 
+RUN apk add --no-cache \
+	netcat-openbsd \
```
pichouk: Some maniac style comment here : can you order package list by alphabetical order and use the same indent please ? :)

app/Dockerfile
```diff
@@ -25,6 +30,7 @@ RUN mkdir -p /mattermost/data \
 
 # Configure entrypoint and command
 COPY entrypoint.sh /
+RUN chmod +x /entrypoint.sh
```
pichouk: Useless entry in my opinion. We just need to set the correct permission for the file on the repository, there is no need to do it on Dockerfile I guess.

app/entrypoint.sh
```diff
@@ -78,4 +78,4 @@ if [ "$1" = 'platform' ]; then
   echo "Starting platform"
 fi
 
-exec "$@"
+exec "$@"
```
pichouk: Please keep newline at the end of the file :)

pichouk:
I really **really** want to use Alpine for Mattermost Docker image, but this is an important thing and we cannot decided alone.
There is an official list of [supported operating systems](https://docs.mattermost.com/install/requirements.html#server-software), but Alpine is not in that list. I think we should ask Mattermost staff to discuss if they could consider to support Alpine officially and/or if it sounds ok for them to use it on our Docker image.

@jasonblais I would appreciate your opinion on this. If there is also some developers at Mattermost with an experience on Alpine, I'll like to hear from them too :)

andruwa13:
@pichouk all fix

pichouk:
Great, good job 👍 Thanks !

pichouk:
Tests failed but it seems that's because you change `entrypoint.sh` mode from 755 to 644.
Just `chmod +x entrypoint.sh` and commit the change please :)

app/Dockerfile
```diff
@@ -29,6 +29,7 @@ RUN mkdir -p /mattermost/data \
 
 # Configure entrypoint and command
 COPY entrypoint.sh /
+RUN chmod 750 /entrypoint.sh
```
pichouk: No, I think we should not use this instruction on the Dockerfile. I explained myself badly I guess.
Just `chmod +x entrypoint.sh` inside your Git repository and commit this change :)

andruwa13: fix

app/Dockerfile
```diff
@@ -25,6 +29,7 @@ RUN mkdir -p /mattermost/data \
 
 # Configure entrypoint and command
 COPY entrypoint.sh /
+RUN chmod +x entrypoint.sh
```
pichouk: Same here.
Should remove this line, run `chmod 755 entrypoint.sh` inside your repository and commit the change :)

app/Dockerfile
```diff
@@ -25,6 +29,7 @@ RUN mkdir -p /mattermost/data \
 
 # Configure entrypoint and command
 COPY entrypoint.sh /
+RUN chmod 755 entrypoint.sh
```
pichouk: Still not what we should do. Access mode should be directly on the repository, not an extra-step inside the Dockerfile.
Just remove this line, apply this command on your repository `chmod 755 entrypoint.sh`, commit and push changes :)

pichouk:
TODO/reminder before merge:
- Fix access mode on `entrypoint.sh`
- Test in a running server (@pichouk)
- Update official Mattermost documentation to add Alpine on supported OS (https://docs.mattermost.com/install/requirements.html#software-requirements)

app/Dockerfile
andruwa13: I did not understand what to do, remove chmod 755 entrypoint.sh?

pichouk: Yes remove this line. But it will then only work if correct mode is apply on `entrypoint.sh` file of the repository.
It worked like that before, but it seems you changed it in your PR.  When I check at your changes I see this : 
capture d ecran_2017-11-25_17-40-15
It means that you change mode for the `entrypoint.sh` file from `755` to `644`. You just have to run the command `chmod 755 entrypoint.sh` inside your shell and push changes to fix this.

pichouk: Good 👍

pichouk:
Just test it on my personal server. Seems to work :)

I'll wait for @xcompass review and the [Mattermost doc update](https://github.com/mattermost/docs/pull/1656). Then it is ok to merge for me.

pichouk:
I just tested it on my existing server, it seems to works perfectly ;)
LGTM, I'll wait for @xcompass review and if this is ok, we will merge :)

jasonblais:
@pichouk Before we merge the Alpine changes, there was a proposal for one of the Mattermost members do a quick sanity test as well. We have a ticket for it here https://mattermost.atlassian.net/browse/PLT-8291

pichouk:
@jasonblais Good idea ! :)

app/Dockerfile
```diff
@@ -7,15 +7,19 @@ ENV MM_VERSION=4.4.1
...
+RUN apk add --no-cache \
+	bash \
+	ca-certificates \
+	curl \
+	jq \
+	libc6-compat \
+	libffi-dev \
+	linux-headers \
+	netcat-openbsd \
+	&& rm -rf /var/cache/apk/* \
```
ccbrown: The `--no-cache` flag shouldn't leave anything behind here, so I'm pretty sure this line is unnecessary.

xcompass:
@pichouk did you hear back from MM developers on the support for Alpine?

app/Dockerfile
andruwa13: @ccbrown is this?
`rm -rf / var / cache / apk / * \`

ccbrown: I'm suggesting removing that line:

```
RUN apk add --no-cache \
	bash \
	ca-certificates \
	curl \
	jq \
	libc6-compat \
	libffi-dev \
	linux-headers \
	netcat-openbsd \
	&& rm -rf /tmp/* 
```

> As of Alpine Linux 3.3 there exists a new --no-cache option for apk. This avoids the need to use --update and remove /var/cache/apk/* when done installing packages.

– https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

app/Dockerfile
```diff
@@ -1,21 +1,24 @@
...
-RUN apt-get update \
-    && apt-get -y --no-install-recommends install \
-      curl \
-      jq \
-      netcat \
-      ca-certificates \
-    && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin
 
+RUN apk add --no-cache \
+	bash \
```
pichouk: If you use `/bin/sh` for the `app/entrypoint.sh` script then there is no need to install `bash` package I guess.

pichouk:
LGTM now.
I'm waiting for Mattermost devs to [test this image](https://mattermost.atlassian.net/browse/PLT-8291).
@jasonblais I cannot comment or follow the Jira task. Can you:
- add a link to this PR on the task. I think it will not have changes anymore, you can test.
- ping me when the task is completed on your side

Thanks :)

jasonblais:
@pichouk Sure thing. I've added a link to this PR. I've also referenced the Docker Compose v3 PR https://github.com/mattermost/mattermost-docker/pull/195

pichouk:
Please also add `xmlsec1` package as requested on https://github.com/mattermost/mattermost-docker/issues/225

andruwa13:
@pichouk done

pichouk:
Good. I'll wait for the Mattermost staff to test and validate this change.
In my side I will test again the image on my server, to validate that everything is ok. I invite everyone who can to do the same :)

Also, I'll squash all commits to get a clean git history :)

pichouk:
I squash the commits and test the image. Looks good to me :)
If anyone else want to test, go ahead.
@jasonblais I'll wait for validation from Mattermost team ?

jasonblais:
I think @coreyhulen will be testing it soon :) 

Huge thanks @pichouk and everyone else involved here!

coreyhulen:
Looks good, I've been testing it for a bit.  I want to test for a little longer, but we should be gtg by tomorrow.

Unrelated to this PR, but looks like I hit the issue described at https://github.com/docker/for-mac/issues/2396 since we map `- /etc/localtime:/etc/localtime:ro`

pichouk:
Thanks to @andruwa13 who did the job :)

@coreyhulen It seems that there is a workaround by using an environment variable (`TZ` ?) instead of mounting `/etc/localtime` inside the container. Maybe we can use that method.

pichouk:
Everything looks great, I just deployed it to my production server and everything is fine. Let's merge 🎉 😄 
Thanks again @andruwa13 for your job !

ionbasa:
Are you including Bash in the alpine package? Or just Sh? It seems Mattermost needs/prefers bash?
Maybe worth seeing if it fixes this issue: #234

Just add the following to the dockerfile for the app:
RUN apk add --no-cache bash and change the entrypoint to be bin\bash

@pichouk

PR #218, opened 2017-12-30
Change the working dir for plugins to work

### LINKED ISSUE #217: Made Plugins works
As [some people](https://forum.mattermost.org/t/a-problem-with-plugins-start-up-with-docker-installation/4263/5) have noticed, Plugins are not working well.

At `app` container startup, there is the following error:
```
[2017/12/19 19:53:14 CET] [EROR] failed to start up plugins: mkdir ./client/plugins: no such file or directory
```
Current workaround is to disable Plugins.

We should investigate on how Plugins works, what they need (eg. a `./client/plugins` folder), if it change our default `config.json` file (eg. to enable or disable by default), etc. before it is still a beta feature ;)

Some links to help:
- [Mattermost Plugin documentation](https://docs.mattermost.com/administration/plugins.html)
- [Forum post to track Plugin related issues](https://forum.mattermost.org/t/mattermost-plugins-in-beta/4123/3)

LordVeovis:
Fix #217 with the default configuration on the PluginSettings as they use relative paths.

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -26,7 +26,7 @@ RUN mkdir -p /mattermost/data \
 # Configure entrypoint and command
 COPY entrypoint.sh /
 ENTRYPOINT ["/entrypoint.sh"]
-WORKDIR /mattermost/bin
+WORKDIR /mattermost
 CMD ["platform"]
 
 # Expose port 80 of the container


pichouk:
Correct workaround. Made some test too and it works for me. It should not change anything for users.
I [updated the Mattermost documentation](https://github.com/mattermost/docs/pull/1709) according to this change.

Thanks you :)

pichouk:
Find the issue ! In the `entrypoint.sh` file there is some instructions which assume that the command is `platform` (lines  [17](https://github.com/mattermost/mattermost-docker/blob/master/app/entrypoint.sh#L17) and [20](https://github.com/mattermost/mattermost-docker/blob/master/app/entrypoint.sh#L20)).
There are multiples way to fix this, but the easiest is to not change the `CMD` instructions inside the Dockerfile, and keep `CMD ["platform"]`. It will work because [line 4 of the Dockerfile](https://github.com/mattermost/mattermost-docker/blob/master/app/Dockerfile#L4) added `/mattermost/bin` to `PATH` var.

I just fixed the PR, it works now. Thanks @LordVeovis :)

PR #219, opened 2018-01-04
[FIX] URLEncode PSQL Password

lasley:
* URL Encode the Postgres password, allowing for special characters in the connection string.

This uses Python to do the quoting, which is available in Ubuntu by default. If #208 is being considered, we'll either need to install Python in that or adapt the implementation here.

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -58,7 +58,9 @@ if [ "$1" = 'platform' ]; then
   if [ -z "$MM_SQLSETTINGS_DATASOURCE" ]
   then
     echo -ne "Configure database connection..."
-    export MM_SQLSETTINGS_DATASOURCE="postgres://$MM_USERNAME:$MM_PASSWORD@$DB_HOST:$DB_PORT_NUMBER/$MM_DBNAME?sslmode=disable&connect_timeout=10"
+    # URLEncode the password, allowing for special characters
+    ENCODED_PASSWORD=$(printf %s $MM_PASSWORD | jq -s -R -r @uri)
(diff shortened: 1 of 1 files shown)


app/entrypoint.sh
```diff
@@ -58,7 +58,11 @@ if [ "$1" = 'platform' ]; then
   if [ -z "$MM_SQLSETTINGS_DATASOURCE" ]
   then
     echo -ne "Configure database connection..."
-    export MM_SQLSETTINGS_DATASOURCE="postgres://$MM_USERNAME:$MM_PASSWORD@$DB_HOST:$DB_PORT_NUMBER/$MM_DBNAME?sslmode=disable&connect_timeout=10"
+    # URLEncode the password, allowing for special characters.
+    if [ "${MM_PASSWORD}" ]; then
+        ENCODED_PASSWORD=$(python -c "import urllib; print urllib.quote('''$MM_PASSWORD''')")
+    fi
```
ccbrown: Since we use `jq`, you can do this more simply with it: `printf %s $MM_PASSWORD | jq -s -R -r @uri`

And that'll work on Alpine as well.

lasley:
Thanks for the review @ccbrown - I altered the implementation as suggested & it works well:

```
/ # echo $MM_PASSWORD
)(*&^%$#@\!
/ # printf %s $DB_HOST | jq -s -R -r @uri
)(*%26%5E%25%24%23%40%5C!
/ # ENCODED_PASSWORD=$(printf %s $MM_PASSWORD | jq -s -R -r @uri)
/ # echo $ENCODED_PASSWORD
)(*%26%5E%25%24%23%40%5C!
```

pichouk:
Good idea, thanks @lasley (and @ccbrown for your review :) )

PR #220, opened 2018-01-07
Auto SSL creation and renewal

akrmhrjn:
Hi,

I have added the code to automatically generate the SSL certificate from Let's Encrypt during docker-compose up. This might be useful as it is very easy to set up. I have also added what needs to be done to enable the automatic generation of SSL certificate. It also checks for SSL certificate expiration and automatically renews the SSL certificate.

Please review it and Let me know if I can be more useful.

Regards.


web/Dockerfile
```diff
@@ -1,11 +1,23 @@
-FROM nginx:mainline-alpine
+FROM nginx
```
xcompass: Any chance make it work with Alpine image?

web/Dockerfile
```diff
@@ -1,11 +1,23 @@
...
+
+# install cron, supervisor and certbot
+RUN echo "deb http://ftp.debian.org/debian stretch-backports main" | tee -a /etc/apt/sources.list
+RUN apt-get update
+RUN apt-get install -qy cron supervisor python-certbot-nginx -t stretch-backports
```
xcompass: Is it possible to use a separate container instead of supervisor?

web/Dockerfile
akrmhrjn: I can try to make work on alpine. Let me take a look on that.

web/Dockerfile
akrmhrjn: Do we need an extra container?

J0WI:
Please don't add unnecessary dependencies to the nginx container. There is already a [Certbot](https://github.com/certbot/certbot) container that can be used for it.

akrmhrjn:
@pichouk Another container seems to be the good option to include LE and manage certificate.

And I agree with replacing web image with Traefik. It would be a good enhancement. It is minimal and has LE support as well.

lim-daniel:
Try use this with matermost:
https://github.com/jwilder/nginx-proxy
https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion

Issue #203, opened 2017-11-19
Use Docker volumes


yukiisbored:
With docker volumes, you can create a virtual volume for bindings. This is useful to save stateful data such as configuration. The reason why using docker volumes is better than binding it to host is to isolate it more from the host and make it more friendly in swarm configurations.

pichouk:
I think we should keep Host mounted volumes on the `docker-compose.yml` file, it could be a breaking change to lot of people if we change this behavior. Also, it allow an easier direct access to data than Docker volumes.

But I agree it could be interesting to use Docker volumes with Swarm. If you submit a PR to add this to the [`contrib/swarm/docker-stack.yml`](https://github.com/mattermost/mattermost-docker/blob/master/contrib/swarm/docker-stack.yml) file, I'll be happy to review and merge :)

yukiisbored:
@pichouk We can just have another docker-compose file like `docker-compose.v3.yml` which leverages the features v3 and Docker 1.10.3+ have.

pichouk:
@xcompass What do you think about this ?
I didn't like the idea to have multiple `docker-compose.vX.yml` file because it can confused people. But maybe I'm wrong about this and it could be better.

xcompass:
Just did a reading on [docker volume](https://docs.docker.com/engine/admin/volumes/). They recommend using volume instead of binding for storing data. I don't have any objection to switch to volume. However, we need to figure out a way for current users to migrate with proper documentation.

HorayNarea:
In favour of bindings: they are easier to backup

pichouk:
Moving to Docker volumes is a big changes for current users and I kind of like bindings since it is easy to backup/modify your data (eg. Mattermost `config.json` file).
But in the other hand, Volumes (and generally Swarm/Docker compose 3 features) seems to be more and more in demand.

I suggest to officially support Docker Swarm on this repository and move `contrib/swarm/docker-stack.yml` at the root of this repository. Then we'll be able to keep some "old-school" stuff inside `docker-compose.yml` and add some Swarm oriented features (like using Docker volumes) on `docker-stack.yml`. As soon as we add correct documentation about, seems ok for me.

Seems good for you ?

yukiisbored:
@pichouk I agree, sounds like a good idea.

pichouk:
I keep this issue opened but in standby for the moment. We will probably change our repository organization with PR#195. We will probably add Docker volumes features to the new Swarm-oriented configuration file.
[...]

PR #222, opened 2018-01-15
Swarm file flavors

ulm0:
After reading the swarm file you guys offer, i saw it could be improved by "un-exposing" some ports for services that can talk to each other through the overlay network, and by adding some more documentation at the top of the file.


contrib/swarm/docker-compose.yml
```diff
@@ -0,0 +1,99 @@
...
+    volumes:
+      # This directory must have cert files
+      - /var/lib/mattermost/cert:/cert:ro
+      - /etc/localtime:/etc/localtime:ro
+    environment:
+      # use app service's hostname
+      - APP_HOST=mm_app
+      # talk to the port within the overlay network
+      # without (over)exposing ports
+      - APP_PORT_NUMBER=80
```
ulm0: Did some digging in the app container and it doesn't listen on port 8065 as stated in the current swarm file, but the port 80.

contrib/swarm/docker-compose.yml
```diff
@@ -0,0 +1,99 @@
...
+      - MM_DBNAME=mattermost
+      # pass the edition to be used, default is enterprise
+      # setting this env var will make the app use the team edition
+      - edition=team
+      # in case your config is not in default location
+      # - MM_CONFIG=/mattermost/config/config.json
+    deploy:
+      restart_policy:
+        condition: on-failure
+  web:
```
ulm0: I've also been able to make mattermost work without the web container provided here, and using traefik instead. Could add an example if wanted.

contrib/swarm/docker-compose.yml
pichouk: Good point, don't know why it was like that...

contrib/swarm/docker-compose.yml
pichouk: I'm okay to drop this container on the stack file [as we discussed here](https://github.com/mattermost/mattermost-docker/issues/203). I use (and prefer) Traefik too.

pichouk:
Looks good to me, waiting for @xcompass approval too :)

contrib/swarm/docker-compose.yml
ulm0: As mentioned in that thread, this repo may keep flavors with the stack file, one with this web image, another with traefik.

ulm0: As mentioned in that thread, this repo may keep flavors with the stack file, one with this web image, another with traefik.

ulm0:
Btw i renamed the file `docker-compose.yml`. should i name it `docker-stack.yml` instead?

contrib/swarm/docker-compose.yml
pichouk: If we can keep both that's a good things, but I don't know how to do that without duplicating a lot of configuration on both files ?

pichouk:
I prefer to keep `docker-stack.yml` unless you have a good argument to use `docker-compose.yml` ?

contrib/swarm/docker-compose.yml
ulm0: Ya, the config duplication is a problem, but I think maintaining both, the current one and the one with traefik will be enough support.

pichouk: Ok, good for me to get both is you want to provide a Traefik example :)

ulm0:
Nah, I just use `docker-compose.yml` for convention, since it uses the Compose file version 3 reference, although is aimed for swarm only, but i can name it back to `docker-stack.yml` in order to avoid confusion.

ulm0:
@pichouk, just added the tfile with traefik

pichouk:
LGTM, thanks @ulm0 :)

PR #227, opened 2018-01-20
Fix broken link

kaakaa:
Fix 404 broken link.

---

There is something I want to confirm.

This fixing and [More Information](https://github.com/mattermost/mattermost-docker#more-information) in README are pointing mattermost docs about ubuntu-**14.04**.
Is it right? should be point ubuntu-**16.04**?

### DIFF
--- a/web/mattermost-ssl
+++ b/web/mattermost-ssl
@@ -55,4 +55,4 @@ server {
     }
 }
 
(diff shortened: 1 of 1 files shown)


pichouk:
Hi ! Thanks, that's a good point !

Yes, if you can change to point on `ubuntu-16.04` I agree :)

PR #235, opened 2018-02-11
updating link about install instruction for Ubuntu

kaakaa:
Updating link about the install instruction for Ubuntu from `14.04` to `16.04`.
Be same as the other page(refs #227).


pichouk:
Thanks 👍

PR #237, opened 2018-02-16
update dockerfile to use MM 4.7.0

cpanato:


pichouk:
@andruwa13 already upgraded on #236 but you also updated the contrib Kubernetes Deployment, so merging too :)
Thanks !

Issue #242, opened 2018-03-05
deactivate TLSv1 by default


cht47:
Hi, 
today I've seen that TLS 1.0 and 1.1 is on by default under web/mattermost-ssl
`ssl_protocols TLSv1 TLSv1.1 TLSv1.2;`

I recommend to reject at least TLSv1. Would be nice if we could edit it in the docker-compose.yml.

pichouk:
I'm okay to remove TLS1.0 since it seems that TLS1.1 is widely supported. If you have some others suggestions to improve TLS security on the `web` image I'll be happy to hear about them.

Would you like to submit a PR ? If yes, fell free to submit. Otherwise I can make the change.

cht47:
Thanks, but you can make the change.

As I don't use the standard HTTPS port [SSLLabs](https://www.ssllabs.com/ssltest/) doesn't work for me to check for further vulnerabilities. 

A recommendation (for a TLS1.2 config): 
`ssl_ciphers ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-SHA;`
There is a huge explanation why these and why this order. Mozilla gives some information about this too.
[Mozilla](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility)
and here you have a generator:
[https://mozilla.github.io/server-side-tls/ssl-config-generator/](https://mozilla.github.io/server-side-tls/ssl-config-generator/)

Enabling HSTS would also be something to think about it. But that can be annoying with cert problems so I don't recommend it for this project.

cht47:
a little update from my side. I will update my container to 5.0 soon and then test some "casual" security configurations without TLSv1 and then some stronger configurations. Then I test it some weeks (colleagues will be my Guinea pigs) and then you will hear from me again with a proper tested working config.

minj:
security > interoperability

drop everything but v1.2, let the unfortunate souls misconfigure their servers on their own

J0WI:
TLSv1.3 support has just been added to nginx on Alpine.

pichouk:
We will probably deprecate the `web` image with #366

PR #243, opened 2018-03-08
Improve TLS security

pichouk:
Some improvements proposal from @cht47 on #242.

I'll appreciate review since I'm not an expert with TLS recommendations :)

### DIFF
--- a/web/mattermost-ssl
+++ b/web/mattermost-ssl
@@ -16,8 +16,8 @@ server {
     ssl_certificate /cert/cert.pem;
     ssl_certificate_key /cert/key-no-password.pem;
     ssl_session_timeout 5m;
-    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-    ssl_ciphers HIGH:MEDIUM:!SSLv2:!PSK:!SRP:!ADH:!AECDH;
+    ssl_protocols TLSv1.1 TLSv1.2;
+    ssl_ciphers ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-SHA;
     ssl_prefer_server_ciphers on;
 
     location ~ /api/v[0-9]+/(users/)?websocket$ {


pe-pan:
I've just cloned and build mattermost-docker and I'm unable to open the UI using HTTPS neither in Chrome Version 64.0.3282.186 (Official Build) (64-bit) nor in Mozilla FireFox 54.0.1 (32-bit).

Chrome gives me:

_This site can’t provide a secure connection
mattermost.hcm.demo.local uses an unsupported protocol.
ERR_SSL_VERSION_OR_CIPHER_MISMATCH
Unsupported protocol
The client and server don't support a common SSL protocol version or cipher suite._

The issue did not get fixed even when upgrading Chrome to Version 65.0.3325.146 (Official Build) (64-bit) or FF to 58.0.2 (64-bit).

I think we need to revert this commit.

cht47:
Can you change it to TLS1.2 only and test it again? My suggestion was only for 1.2 and without warranty why I posted the Mozilla link. I build one for 1.2
`    ssl_protocols TLSv1.2;
    ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256';
    ssl_prefer_server_ciphers on;
`
For old browsers
` ssl_protocols TLSv1.1 TLSv1.2;
    ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
    ssl_prefer_server_ciphers on;`

pe-pan:
Ok, this works for the newest browsers; but I don't see the point - you want to support TLS1.2 only? 

Also, I'm not sure why to remove the support of TLS 1. I know it's not secure but shouldn't this be up to the client to decide which security version to use? What if they just don't have browser new enough... IMHO, MM server should support all the security versions and client should decide itself...

cht47:
https://www.acunetix.com/blog/articles/tls-vulnerabilities-attacks-final-part/

So your suggestion is to just give the lowest possible security to support the 90's and if the client wants more security it just can ask for it? What do you do if a hacker pretends to support only TLS 1.0 or even SSL and uses its vulnerabilities to hack you.

Most people use the Mattermost client anyway which should work with TLS1.2. I don't think that a company which uses modern container services like Docker, uses on the other hand some very old browsers.

Here is a list https://help.salesforce.com/articleView?id=000220586&language=en_US&type=1
TLS1.2 Support since Chrome v30, FF v27 and IE 11.

pe-pan:
I'm not here to analyze the issue and provide the solution. I'm just claiming the recent changes are not working in Chrome / FF. Either I'm missing some settings or the changes were wrong. Please, either explain which settings I'm missing or provide proper fix or rollback the changes.

pichouk:
My bad, I think I should not have merge this without another review since I'm not expert in TLS configuration.
I revert the commit for the moment to be sure it will not be an issue for users. **BUT** I strongly agree that security should be a default configuration, not an option.

So @cht47 if you can help me on https://github.com/mattermost/mattermost-docker/issues/242 to understand and find a configuration which can be a good balance between security and compatibility I'll be happy.

Also, I am on this project for few months now and I realized that the most complicated component to maintain is the Web container. Why ? Because there is **A LOT** of different needs for a reverse-proxy, everybody need a specific configuration (Let's Encrypt, TLS compatibility, etc.). So it will lead us to build a big image with a lot of options to customize Web container, only to propose a basic reverse-proxy (which is not the main goal of this repository IMHO). That's why I opened a [Github issue](https://github.com/mattermost/mattermost-docker/issues/245) to discuss about the future of this image (and I invite you to react on this issue). My opinion is that users should use their own reverse proxy and we don't need to propose and maintain a basic Nginx image.

PR #261, opened 2018-04-18
Update docker-compose.yml for new port

spinanicky:
With the change to 8000 the port on the web proxy needs to be updated

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,7 +44,7 @@ services:
   web:
     build: web
     ports:
-      - "80:80"
+      - "80:8000"
       - "443:443"
     read_only: true
     restart: unless-stopped


pichouk:
Hi @spinanicky. Thanks for your help but the port that changed is the Mattermost application port, not the port of the web server.

The web container is [still listening on port `80`](https://github.com/mattermost/mattermost-docker/blob/master/web/mattermost#L7) and if forwarding, by default, to the app container [port `8000`](https://github.com/mattermost/mattermost-docker/blob/master/web/entrypoint.sh#L5). So everything is fine, no needs to change the `docker-compose.yml` :)

spinanicky:
@pichouk - You are correct - my apologies. I had an error due to use of 443.

PR #305, opened 2018-09-10
Use latest alpine image for app container

### LINKED ISSUE #304: Update Alpine base
These are both based to an outdated version of the upstream image:
https://github.com/mattermost/mattermost-docker/blob/17b9055ed66d03361939cf88fb1f44da498880b6/app/Dockerfile#L1
https://github.com/mattermost/mattermost-docker/blob/17b9055ed66d03361939cf88fb1f44da498880b6/db/Dockerfile#L1

The web image is based on the latest tag:
https://github.com/mattermost/mattermost-docker/blob/17b9055ed66d03361939cf88fb1f44da498880b6/web/Dockerfile#L1
This might be suitable for Alpine and Postgres too (e.g. `postgres:9-alpine`).

tejasbubane:
Closes #304

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.6
+FROM alpine:3.8
 
 # Some ENV variables
 ENV PATH="/mattermost/bin:${PATH}"


app/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.6
+FROM alpine:latest
```
cpanato: I would not use latest. instead please pick the latest tag, if is `3.8` set here.

tejasbubane:
@cpanato Fixed.

PR #315, opened 2018-09-28
if all config was set in the config.json bypass the checks

cpanato:
This makes the image more k8s friendly for our helm charts

if all the config was set in the config.json and we provide that dont need to check the DB. 
also the check for the db in k8s is made by an init container.

this improve this image to work in k8s envs.

cc @jwilander


cpanato:
just test this in our helm chart and works like a charm 😄

cpanato:
@pichouk thanks!!

PR #316, opened 2018-10-01
Remove DB check in the app/entrypoint

cpanato:
just realize that this PR https://github.com/mattermost/mattermost-docker/pull/315 checks the variable DB and db exist always we need to check if a config or the MM_SQLSETTINGS_DATASOURCE not set

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -67,16 +67,6 @@ if [ "$1" = 'mattermost' ]; then
     echo "Using existing config file" $MM_CONFIG
   fi
 
-  if [ -z "$MM_SQLSETTINGS_DATASOURCE" ] && [ ! -f $MM_CONFIG ]
-  then
-    # Wait for database to be reachable
-    echo "Wait until database $DB_HOST:$DB_PORT_NUMBER is ready..."
-    until nc -z $DB_HOST $DB_PORT_NUMBER
-    do
-      sleep 1
-    done
-  fi
-
   # Wait another second for the database to be properly started.
   # Necessary to avoid "panic: Failed to open sql connection pq: the database system is starting up"
   sleep 1
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -67,7 +67,7 @@ if [ "$1" = 'mattermost' ]; then
     echo "Using existing config file" $MM_CONFIG
   fi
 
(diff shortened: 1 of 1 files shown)


app/entrypoint.sh
```diff
@@ -67,7 +67,7 @@ if [ "$1" = 'mattermost' ]; then
     echo "Using existing config file" $MM_CONFIG
   fi
 
-  if [ ! -z "$DB_HOST" ]
+  if [ -z "$MM_SQLSETTINGS_DATASOURCE" ] && [ ! -f $MM_CONFIG ]
```
pichouk: Don't sure if it will works. If the file `MM_CONFIG` does not exists it will be created before in this script (see line 30), so I guess this condition will never be true.

cpanato: i tested all the cases here
1. with a mm config
2. without mm config but the MM_SQLSETTINGS_DATASOURCE set
3. without mm config and without MM_SQLSETTINGS_DATASOURCE set

pichouk: This is weird. If there is no configuration file it will be created by [this line](https://github.com/mattermost/mattermost-docker/blob/b57346b185e530974ce6ebcc6e8e9b1e01250ac1/app/entrypoint.sh#L36) so, at this point, a configuration file should always exists.

cpanato: can you try on your side?

pichouk: I'll do it later to check yes.

cpanato: thanks @pichouk

pichouk: I confirm on my side, after some test, that this condition is never `true` when using a "basic" Docker Compose install. Container is correctly working, but the lines under this condition are here for safety, to prevent Mattermost application to boot if the database is not reachable. We have 2 options :
- remove this check of the database and consider that if database is not available then application will not boot anyway
- fix this condition to match your case and the "basic" one

cpanato: Thanks for checking.
I agree to remove the check. This is performed by the app.

Agree to remove @pichouk

pichouk: Yes LGTM :+1:

cpanato: ok, removed PTAL

pichouk:
Perfect, thanks @cpanato :)

PR #324, opened 2018-10-14
fix S3 wal-e support

cobenash:
I made some simple adjustments so that wal-e can be reused. Check Please.

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -19,7 +19,7 @@ RUN apk add --no-cache \
 COPY setup-wale.sh /docker-entrypoint-initdb.d/
 
 #Healthcheck to make sure container is ready
-HEALTHCHECK CMD pg_isready || exit 1
+HEALTHCHECK CMD pg_isready -U $POSTGRES_USER -d $POSTGRES_DB || exit 1
 
 # Add and configure entrypoint and command
 COPY entrypoint.sh /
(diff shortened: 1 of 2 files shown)


db/Dockerfile
```diff
@@ -19,11 +19,11 @@ RUN apk add --no-cache \
 COPY setup-wale.sh /docker-entrypoint-initdb.d/
 
 #Healthcheck to make sure container is ready
-HEALTHCHECK CMD pg_isready || exit 1
+HEALTHCHECK CMD pg_isready -U postgres || exit 1
```
pichouk: I remove the user in PR #318 because the user is not always postgres. So the healthcheck return a FATAL log message.

pichouk:
I'm not sure, I don't see the link between those 2 PR. I guess the script cannot write to `/etc/wal-e.d/env/` because it is started as `read_only`. I don't know if there is a workaround for this or if we need to remove the `read_only` setting.

cobenash:
@pichouk,
Yes, I add this folder to volume at the bottom of the DB Dockerfile. It work again with this settings.

 Could you please test this PR again?  If there are something wrong, please let me know.

pichouk:
I restarted the Travis build several time and there is still the same error. I don't know if adding the folder to the `VOLUME` directive should have work or not...

cobenash:
@pichouk I found out the problem. The main problem is setup-wale.sh file not volume I added at DB dockerfile.

I used the environment AWS_ACCESS_KEY_ID ... variable as an input but it was empty at the travis CI process. 

I will do some modify and update the branch.

pichouk:
Tests are OK now, is the PR OK for you ?

cobenash:
@pichouk Yes. :)

Issue #245, opened 2018-03-11
Drop support for Web image ?


pichouk:
I am thinking of dropping support and remove the web image from this repository, but I really think this is a decision to be discussed, so please give me feedback on this (whether you **agree or not**).

Actually, I think that the Web image is not really useful since it is only a Nginx server with minimal configuration to be used as a reverse proxy. I'm pretty sure that a huge majority of users already have their own reverse-proxy inside their infrastructures, so they are not using Web image. And people who don't should really, IMHO, use an existing reverse-proxy (Traefik, Nginx, etc.) and customize their configuration instead of using the default one we provide.

So what are you thinking ? Should we stop providing a Web image (and instead explain in documentation how to use App with a reverse-proxy) ? Or should we keep this image and improve it (maybe by using another reverse-proxy, like Traefik) ?

HorayNarea:
I think having a default, small, "security-first" reverse proxy is a good thing… but more as an example and "how to get this up and running as secure and fast as possible" not as a solution for everybody

jasonblais:
@pichouk What are the implications of dropping support for the web image? What harm is keeping the web image support?

pichouk:
@jasonblais Dropping support for Web image will force users that use this image (how much ? can't really know) to configure their own reverse-proxy/web server **OR** to expose directly the Mattermost application to the internet.  
IMHO this is not something difficult if users have some basic system administration skills (which I hope they have since they host a Mattermost server), but we will need to warn them (it will not be transparent if they just `git pull && docker-compose up`).

The complicated task with maintaining the Web image is "choices". A reverse-proxy/web server is the "main door" to enter on someone's infrastructure so this is a sensitive and highly customized component. There is probably hundreds of possible reverse-proxy/web server and thousands of different configurations. Somes [want specific features](https://github.com/mattermost/mattermost-docker/issues/162), others [want light and generic](https://github.com/mattermost/mattermost-docker/pull/220#issuecomment-366546279) image. Somes [want security](https://github.com/mattermost/mattermost-docker/issues/242), others [want compatibility](https://github.com/mattermost/mattermost-docker/pull/243#issuecomment-372372754).
There are as much possible configurations as there are users. Because of this, it is really complicated to provide an image that will fit everyone's needs, and I guess that a lot of users don't use Web image for this reason (and use their own reverse-proxy).

jasonblais:
[...]

PR #348, opened 2018-12-12
Allow acme challenge through NGINX configuration

jfbrazeau:
In docker production, the SSL certificate is enabled through NGINX configuration. It would be convenient to ease the integration of *let's encrypt*  certificates by allowing to expise acme challenge files through HTTP (`http://host:port/.well-known/acme-challenge/<challenge file>`).

This merge request is a proposal to enable this through an additional docker volume for the web component (`./volumes/web/acme-challenge`).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,7 @@ services:
       # This directory must have cert files if you want to enable SSL
       - ./volumes/web/cert:/cert:ro
       - /etc/localtime:/etc/localtime:ro
+      - ./volumes/web/acme-challenge:/var/www/acme-challenge:ro
     # Uncomment for SSL
     # environment:
     #  - MATTERMOST_ENABLE_SSL=true
--- a/web/mattermost
+++ b/web/mattermost
@@ -36,4 +36,10 @@ server {
         proxy_read_timeout 600s;
         proxy_pass http://{%APP_HOST%}:{%APP_PORT%};
     }
+
+    location /.well-known/acme-challenge/ {
+        default_type "text/plain";
+        root         /var/www/acme-challenge;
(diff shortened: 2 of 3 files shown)


pichouk:
Hi :)

Thanks, this is a good idea. In fact we will deprecate the Web image [one day](https://github.com/mattermost/mattermost-docker/issues/245), but, as you can see, this subject is open for a while now (and it's my fault). So I'll accept this PR after few tests, but I just want you to know that we might stop support for the Web image sooner or later ^^

pichouk:
I didn't understand what component request Let's Encrypt for the certificate here ? Is this supposed to work if we enable Let's Encrypt on the Mattermost application ? Because I don't see how it can works since the web container and the app container are not sharing volumes together.

jfbrazeau:
As far as I can see, there are 2 ways to configuring TLS on mattermost server (as we can see here : https://docs.mattermost.com/install/config-tls-mattermost.html ). The first is to setup TLS directly on Mattermost Server as you mention. The second is to configure it through NGINX proxy.

In mattermost-docker, as far as I can see the second solution has been chosen (as you can see here : https://github.com/mattermost/mattermost-docker#install-with-ssl-certificate , the SSL certificate must be registered in NGINX, not mattermost).

As a consequence, I came to the conclusion that it was more relevant to add the acme challenge capacity to the NGINX docker image. But maybe it wasn't a good conclusion ! 😊 

You say that you may stop the support for the web image : may I ask you why you plan to remove it ?

hanzei:
Hey @pichouk,

What are the next steps on this PR?

pichouk:
We will [deprecate and remove](https://github.com/mattermost/mattermost-docker/pull/366) the `web` image so we can close this one.  
If someone wants to have a TLS setup, an example based on Traefik will be add by #366

PR #359, opened 2019-01-29
Break traefik command in multiple lines

Ovski4:
Hello,

The traefik command is pretty long. You can use yaml folded style:
- Each line break is replaced by a space.
- The indention in each line will be ignored
- A line break will be inserted at the end.
Which keep the current behaviour


pichouk:
It is better indeed, thanks @Ovski4 :) I'll merge when you'll sign the Mattermost CLA :)

cpanato:
CLA is signed will merge

cpanato:
@Ovski4 thanks for your contribution!

PR #390, opened 2019-05-05
include timezone for plugins that use location information.

scottleedavis:
There has been [some issues](https://github.com/scottleedavis/mattermost-plugin-remind/issues/88) with running a plugin that uses time.In(location), where location is loaded by tzdata information.  

http://pouzek.si/blog/go-loadlocation-docker/

This PR adds the tzdata apk to the app container

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -22,6 +22,7 @@ RUN apk add --no-cache \
 	mailcap \
 	netcat-openbsd \
 	xmlsec-dev \
+	tzdata \
 	&& rm -rf /tmp/*
 
 # Get Mattermost


scottleedavis:
one of travis builds looks stalled on
```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
```
I don't think that is related to this PR?   (correct me please if wrong)

scottleedavis:
also affect another plugin https://github.com/standup-raven/standup-raven/issues/14

pichouk:
LGTM, good idea, thanks :)

PR #391, opened 2019-05-15
fix typo in readme

redg3ar:


pichouk:
`comment out` is equivalent to `uncomment` right ? So I guess the README is correct. Maybe I'm wrong, I'm not English native speaker.

redg3ar:
Comment out means "remove by commenting" to me. Like, "taking the code out of the file by commenting". Since we are reintroducing code by removing the comment, I think uncomment makes sense here.

pichouk:
Then LGTM, thanks :)

PR #403, opened 2019-07-18
Improve Docker Images and Travis CI

8ear:
Hi Guys,

Is it possible to improve the Dockerfile of App image?
And the repository `https://hub.docker.com/r/mattermost/mattermost-team-edition` does other things as `https://hub.docker.com/r/mattermost/mattermost-prod-app`. With my solution you can use Travis CI for building, testing, retagging and pushing to hub.docker.com.

You need to store an DOCKER_USERNAME and a DOCKER_PASSWORD protected Travis CI environment variable in the project.

The following improvements are made:
- PUID and PGID are flexible so if the mattermost user reuqire to have another UID for bind paths for example it is possible
- I updated Travis CI to use the newest xenial infrastructure
- I added a hostname to Travis CI
- Make the MM_VERSION variable flexibel, so you can use the image with all tags you require.
- Renamed MM_USERNAME and MM_PASSWORD variables to MM_DB_USERNAME and MM_DB_PASSWORD. So that you have a clear variable scheme. All variables starts with MM and followed her next use case. So I also renamed the MM_DBNAME to MM_DB_NAME
- I added the su-exec package, so every command after the entrypoint are executed as mattermost user. But in the entrypoint you can make root tasks as well.
- Quoted all variables.
- Added shellcheck for all *.sh shellscripts

If you want any other please tell me or change it self (you have the right :-) ).
I does also see your changes for Version 6 they are great. I think also that docker-compose file should only show an example. But then the image should be ready at hub.docker.com for enterprise and team edition with a few possible modifications and a complete docker-compose and docker run example is required at the hub.docker.com readme.

I will use mattermost in my big docker-compose stack file at rancherOS (which rancher user id is 1100).

Kind regards


jasonblais:
Hey @cpanato, what do you think about the changes here?

cpanato:
hello @8ear thanks for your PR
We are planning to deprecate this dockerfile in this repo in favor the one we move to the server repo (see: https://github.com/mattermost/mattermost-server/blob/master/build/Dockerfile) so we are not making any changes here related to the docker file. only for the examples to run in the environments.

The images we are generating now is based on the docker file in the server repo

cpanato:
closing due no activitie and we are deprecating this docker file soon.
the new docker file you can find here: https://github.com/mattermost/mattermost-server/blob/master/build/Dockerfile

PR #421, opened 2019-10-22
[MM-18794] mattermost command line broken in Docker since the 5.11.1 -> 5.12.0 transition

cpoile:
#### Summary
- The env variable `MM_SQLSETTINGS_DATASOURCE` exported in `app/entrypoint.sh` isn't kept as an environment variable in the app container after startup. 
- One solution is to manually define it in the `docker-compose.yml` file. This keeps it in the environment, so when the CLI is started sometime later, the correct database connection string is used.
- Totally open to other solutions. 
- I tried to reference the variables (eg, `${MM_USERNAME}`) in the string, but docker-compose isn't set up to do that. One way would be to use an environment file, import it, and then you could reference those variables. Do you want to do that @cpanato?

#### Next Steps:
- This is being felt by customers using 5.12-on, should we update those releases with this?

#### Ticket
- [MM-18794](https://mattermost.atlassian.net/browse/MM-18794)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,6 +40,11 @@ services:
       - MM_USERNAME=mmuser
       - MM_PASSWORD=mmuser_password
       - MM_DBNAME=mattermost
+
+      # use the credentials you've set above, in the format:
+      # MM_SQLSETTINGS_DATASOURCE=postgres://${MM_USERNAME}:${MM_PASSWORD}@db:5432/${MM_DBNAME}?sslmode=disable&connect_timeout=10
+      - MM_SQLSETTINGS_DATASOURCE=postgres://mmuser:mmuser_password@db:5432/mattermost?sslmode=disable&connect_timeout=10
+
       # in case your config is not in default location
       #- MM_CONFIG=/mattermost/config/config.json
 


cpoile:
Please excuse the force push -- I rebased off master.

cpoile:
@gabrieljackson I know, I feel the same way. Nothing exported from the app/entrypoint.sh is around in the container after startup, and it's not just from the if -- I exported something right before the final `exec` call, and it wasn't there. So I'm totally confused about what's happening here.

As for backporting -- I don't know if backporting makes sense in this repo, true, but I'd like to if it does. If this is the solution, I'll update the documentation and then I'll go through the customer threads about this bug and explain what to do.

gabrieljackson:
Sounds good @cpoile. It's really confusing as you said, because everything points to the entrypoint file seemingly being fine the way it is.

Let me review it again to see if I can figure out what is going on.

cpoile:
@cpanato Gentle ping on this -- some customers really want this fixed (see the thread linked in the Jira ticket).

Zoddo:
> Hmm, I just went and reviewed `entrypoint.sh` and I didn't immediately see why the variables and settings exported there are not around after the server starts as the server is called from within that shell file.

I think I see the issue...

The environnement variables exported from `entrypoint.sh` are there... in the server process.
When you use `docker exec`, the `entrypoint.sh` script isn't executed again. Therefore, the command is started from a scratch environment, with only the env defined in the container's config.

While the new process (created from `docker exec`) become a child of the server, it's not a true child (understand: the process wasn't started from the server's process), and therefore doesn't inherit from its environnement. It is set as a child of the server process only because it needs a parent (only the init doesn't have any parent, and the init process is the default parent for orphaned processes).
 
It's a bit like an orphaned process on a "standard" system: because its parent died, it is "relinked" to the init process by the kernel. But that doesn't mean it magically inherits from the init's environnement.

About the solution... If you don't want to manually define `MM_SQLSETTINGS_DATASOURCE` in `docker-compose.yml`, a solution would be to expose in the PATH a `mattermost` wrapper which would export this environnement variable before starting the "true" `mattermost` binary.

cpanato:
looks good to me
I will go over this docker compose and check what errors more we might have, it is ok for now

PR #427, opened 2019-11-19
Bug fixes

sstaszkiewicz-copperleaf:
The postgres base container was stuck in a restart loop. The base entrypoint file was not being called properly, so it exited immediately and postgres never started.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,8 +20,7 @@ services:
     #  - AWS_REGION=us-east-1
 
   app:
-    build:
-      context: app
+    build: app
       # uncomment following lines for team edition or change UID/GID
       # args:
       #   - edition=team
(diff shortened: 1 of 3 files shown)


sstaszkiewicz-copperleaf:
Signed the CLA. Seems like a random 500 in travis, unrelated to these changes.

sstaszkiewicz-copperleaf:
I'm guessing this broke the DB container because of the recent changes to the base container entrypoint.
https://github.com/docker-library/postgres/commit/7c84645f2d38953e1aee1742e8f607ffa9ac5884#diff-4e48159f378a021e74d0939cd5206ef1

Should we lock it down to `9.4.25-alpine` to prevent this kind of break in the future?

theWeird:
The change in db/entrypoint.sh solved the issue for me.

hanneselmer-vendevio:
The changes also fixed my problem, I was not able to start without the change.

docker-compose.yml
```diff
@@ -20,8 +20,7 @@ services:
     #  - AWS_REGION=us-east-1
 
   app:
-    build:
-      context: app
+    build: app
```
C-Duv: Why removing the `context` key?
If one want to use team edition or change UID/GID (as advised below) it will need to re-put the [`context`](https://docs.docker.com/compose/compose-file/#context) to use the [`args`](https://docs.docker.com/compose/compose-file/#args) key.

sstaszkiewicz-copperleaf: Yeah good point. I changed it to follow the pattern from the other services. Also, somebody should have caught that the file permissions were changed from `644 -> 755`.

PR #435, opened 2019-12-17
[Spike] Playing with metrics

esdrasbeleza:
## Summary

This is probably some work in progress that needs more discussion, and it has some proposals to address [MM-21112](https://mattermost.atlassian.net/browse/MM-21112).

I started to write more details here, but since it was becoming a large wall of text, I attached a PDF to the Jira ticket. Sorry for being verbose! :)


esdrasbeleza:
Alright, here we go:

I removed all the components that we decided not to use (Grafana and the Go spikes), and I created a new component/spike, metrics-watcher.

I checked Grafana and Prometheus sources, and turns out Prometheus has a [Golang API](https://github.com/prometheus/client_golang/tree/master/api) that encapsulates the calls to its REST API. It can be used [like this](https://github.com/esdrasbeleza/mattermost-docker/blob/master/metrics-watcher/prometheushelper/sample_test.go#L14-L44).

I [queried the API for two sample metrics](https://github.com/esdrasbeleza/mattermost-docker/blob/c33ee7482b82439ba61ffbf72491269e4deb66a0/metrics-watcher/metrics.go#L5-L23). Once again I'm not sure how useful they are, but they give us some idea of how complex it would be to query some metrics.

Prometheus API is encapsulated inside a [helper](https://github.com/esdrasbeleza/mattermost-docker/blob/master/metrics-watcher/prometheushelper/prometheushelper.go#L17-L53) I created as well. This current implementation only works for vectors whose only useful value is the first one.

The source code has more fun stuff, like querying Prometheus health, tests etc.

## Running locally

`docker-compose up` will run everything, including `metrics-watcher`, and show its output.

By default `metrics-watcher` uses `http://localhost:9090` as the Prometheus HTTP server; this configuration can be overridden using the environment variable `MATTERMOST_PROMETHEUS_SERVER=<some other server>`, as in [docker-compose](https://github.com/esdrasbeleza/mattermost-docker/blob/f5ad80be93a66223fd2880676712bee35de5df89/docker-compose.yml#L82).

## Known problems 😞

When we run docker-compose and the server is not up yet, error messages are going to be printed. A possible workaround is to run a healthcheck for the server, and query Prometheus only if the server is up.

streamer45:
Thank you Esdras, nice improvements! :+1:  

I've had a chance to take a deep look at the new code you pushed.  
I can already see some of your work being integrated in our [new](https://github.com/mattermost/mattermost-load-test-ng) load-test repository, which was the goal of this ticket.

In the end, it doesn't look like much code is needed to get a lot of good information and implement a basic real-time monitoring system for a MM instance. Prometheus API really helps in making this process quite easy of course. We are essentially running some queries and evaluating the returned data. This is a great starting point.

Ideally, I'd like to see a couple more important metrics added (I am thinking database store times and api errors) and, most importantly, a way to fire events when these go above or below a pre-configured threshold.

Moreover, I think it would be great if you could make a _code only_ PR (without the docker boilerplate) for your ```metrics-watcher``` in https://github.com/mattermost/mattermost-load-test-ng so I can review more in detail and easily integrate your work in our existing codebase.

Let me know if that sounds good and/or if you need any help. You can reach me as usual on [community](https://community.mattermost.com/).

esdrasbeleza:
Following @streamer45's suggestions, work was moved to another repo and there's [a new PR](https://github.com/mattermost/mattermost-load-test-ng/pull/93) for it. I'm closing this PR now.

Issue #437, opened 2019-12-24
Probably wrong docker-compose file syntax


rajudev:
When I am trying to un-comment the lines for doing a team edition deployment as specified in the [Production Docker install documentation](https://docs.mattermost.com/install/prod-docker.html) as well as in the [README.md](https://github.com/mattermost/mattermost-docker/blob/master/README.md#choose-edition-to-install)

I get the following error
```
$ git clone https://github.com/mattermost/mattermost-docker.git
Cloning into 'mattermost-docker'...
remote: Enumerating objects: 20, done.
remote: Counting objects: 100% (20/20), done.
remote: Compressing objects: 100% (20/20), done.
remote: Total 1643 (delta 7), reused 10 (delta 0), pack-reused 1623
Receiving objects: 100% (1643/1643), 335.02 KiB | 1.09 MiB/s, done.
Resolving deltas: 100% (806/806), done.
$ cd mattermost-docker/
$ emacs docker-compose.yml 
$ git diff
diff --git a/docker-compose.yml b/docker-compose.yml
index e045a30..027d9d2 100755
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,8 +22,8 @@ services:
   app:
     build: app
       # uncomment following lines for team edition or change UID/GID
-      # args:
-      #   - edition=team
+      args:
+        - edition=team
       #   - PUID=1000
       #   - PGID=1000
     restart: unless-stopped
$ docker-compose build
ERROR: yaml.parser.ParserError: while parsing a block mapping
  in "./docker-compose.yml", line 23, column 5
expected <block end>, but found '<block mapping start>'
  in "./docker-compose.yml", line 25, column 7
```
After looking further with previous issues, I came across this Merge Request  #146 
According to it there should be a `context:` block in the [docker-compose.yml](https://github.com/mattermost/mattermost-docker/blob/b9484f444424057434fd8c0a1c1a937154639857/docker-compose.yml#L26)

So I tried adding `context` and check if the `docker-compose build` works. Apparently it did not.
```
$ emacs docker-compose.yml 
$ git diff
diff --git a/docker-compose.yml b/docker-compose.yml
index e045a30..b0b2487 100755
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,9 +21,9 @@ services:
 
   app:
     build: app
-      # uncomment following lines for team edition or change UID/GID
-      # args:
-      #   - edition=team
+      context: app
+        args:
[...]

PR #438, opened 2019-12-26
Fixes for the smooth deployment of Team edition.

rajudev:
### Summary
**Problem:**
For installation of Team Edition, simply commenting out the two lines mentioned in README does not work. Created issue #437 for it.

**Solution:**
After looking at a few blogs and the commit history of the `docker-compose.yml` . I observed that the `context: ` has been deleted by some other PR which was previously present. So I have 
added it back to make team edition deployments work.

**Fixes:**
	- Added the context block in `docker-compose.yml`
	- Added the subsequent changes in the installation instructions in README.md

#### Ticket Link
Probably wrong docker-compose file syntax #437

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,13 @@ The following instructions deploy Mattermost in a production configuration using
 
 If you want to install Enterprise Edition, you can skip this section.
 
-To install the team edition, change `build: app` to `build:` and uncomment out these lines in docker-compose.yaml file:
+To install the team edition, change `build: app` to `build:` and uncomment out these lines in `app:` services block to make it look like below in docker-compose.yaml file:
 ```yaml
-context: app
-args:
-  - edition=team
+app:
+  build:
+    context: app
+    args:
+      - edition=team
 ```
 The `app` Dockerfile will read the `edition` build argument to install Team (`edition = 'team'`) or Enterprise (`edition != team`) edition.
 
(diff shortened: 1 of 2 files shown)


jasonblais:
cc @cpanato for review after holidays (I don't have permissions to add you as a reviewer)

docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
   app:
     build: app
```
ccoenen: it should be noted that 

```
build: app
```

will not work with sub-items. I reverted to my "old" few lines:

```
  app:
    build:
      context: app
      args:
        - edition=team
```

README.md
```diff
@@ -23,8 +23,9 @@ The following instructions deploy Mattermost in a production configuration using
 
 If you want to install Enterprise Edition, you can skip this section.
 
-To install the team edition, uncomment out these lines in docker-compose.yaml file:
+To install the team edition, change `build: app` to `build:` and uncomment out these lines in docker-compose.yaml file:
```
ccoenen: I think the "build:" is easy to miss, it would be a good idea to just add one more line to the example below so that one can see the full change.

ccoenen: essentially adding this as example:
```
  app:
    build:
      context: app
      args:
        - edition=team
```

rajudev: I made the changes as pointed out by you in `README.md` as well as the relevant part in the `docker-compose.yml` 
Review whenever you can
Thanks

Issue #346, opened 2018-12-07
How backups work?


michaelr524:
I've setup the container with s3 access and it backs up wal-e segments to s3.
The README lacks an explanation how these backups work in general.
Is there a place where I can read more about it?
Specifically, the questions that I have right now are:

1. Is it enough to have these wal-e backups or do I need to also setup cron to run base backups?
2. Does the command to cleanup old base backups also cleans old wal-e segments? 
3. What is the procedure to restore?

michaelr524:
Also, trying to run the base backup I get an error about the s3 prefix not specified. It's specified in the docker-compose.yml file. Should I specify it again separately somehow for the backup commands?

```
$ docker exec mattermost_db_1 su - postgres sh -c "/usr/bin/envdir /etc/wal-e.d/env /usr/bin/wal-e backup-push /var/lib/postgresql/data"
wal_e.main   INFO     MSG: starting WAL-E
        DETAIL: The subcommand is "backup-push".
        STRUCTURED: time=2018-12-07T11:15:48.724757-00 pid=15905
wal_e.main   ERROR    MSG: no storage prefix defined
        HINT: Either set one of the --wabs-prefix or --s3-prefix options or define one of the WALE_WABS_PREFIX, WALE_S3_PREFIX, or WALE_SWIFT_PREFIX environment variables.
        STRUCTURED: time=2018-12-07T11:15:48.725017-00 pid=15905
```

michaelr524:
I found a good explanation here: https://jbub.eu/2016/05/16/postgresql-backup-wale-cloud-files/
The db container exposes a volume for the wal-e environment variables (/etc/wal-e.d/env):
https://github.com/mattermost/mattermost-docker/blob/master/db/Dockerfile#L29

michaelr524:
I made some progress, but there is a new problem now:

```
docker exec mattermost_db_1 su - postgres sh -c "/usr/bin/envdir /etc/wal-e.d/env /usr/bin/wal-e backup-push /var/lib/postgresql/data"
wal_e.main   INFO     MSG: starting WAL-E
        DETAIL: The subcommand is "backup-push".
        STRUCTURED: time=2018-12-08T22:12:37.830561-00 pid=63
psql: FATAL:  role "postgres" does not exist
wal_e.main   ERROR    MSG: could not run one or more external programs WAL-E depends upon
        DETAIL: Could not get a connection to the database: note that superuser access
        is required
[...]

PR #443, opened 2020-01-12
Installing missing package pv

DE-mbecker:
I tried to use wal-e as mentioned on dockerhub (see https://hub.docker.com/r/mattermost/mattermost-prod-db). To do the base backup I had to do some additional steps (see https://github.com/mattermost/mattermost-docker/issues/346):
- mount a volume into the database docker container (path in the container is /etc/wal-e.d/env) and put all credentials and the s3 prefix to this folder
- exec into the container and create a superuser (Command: CREATE USER postgres SUPERUSER;)

Now it still fails with the message, that the package *pv* is missing

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -12,6 +12,7 @@ RUN apk add --no-cache \
       python-dev \
       py-pip \
       py-cryptography \
+      pv \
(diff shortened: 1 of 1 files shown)


DE-mbecker:
Hey,

I've signed the CLA. How do I restart the check?

Kind regards,

Marc

PR #444, opened 2020-01-16
bump to 5.19.0

jomaxro:
#### Summary
Mattermost 5.19 was released today, bumping the Dockerfile ENV variable so Docker installs can upgrade.

#### Ticket Link
N/A

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -2,7 +2,7 @@ FROM alpine:3.10
 
 # Some ENV variables
 ENV PATH="/mattermost/bin:${PATH}"
-ENV MM_VERSION=5.18.1
+ENV MM_VERSION=5.19.0
 
 # Build argument to set Mattermost edition
 ARG edition=enterprise


cpanato:
just a reminder, please start using the images from `mattermost/mattermost-enterprise-edition` or `mattermost/mattermost-team-edition` those will replace this one in version 6

jomaxro:
Any details/documentation on the upcoming change? Where are those images, they don't seem to be public GitHub repos?

cpanato:
@jomaxro the images are in docker hub and we are building those since 5.10 release i guess.

you can find the docker file here: https://github.com/mattermost/mattermost-server/blob/master/build/Dockerfile

for example: our k8s operator use those images and our community.mattermost.com is using that as well

There is an issue in this repo saying we will deprecate the image and also there is a post in the docker channel as well

jomaxro:
Will review, thanks!

PR #452, opened 2020-02-20
Removes the deprecated nginx ssl directive (#451)

### LINKED ISSUE #451: nginx: [warn] the "ssl" directive is deprecated
**Problem**: I'm getting the following warning when starting `mattermost-docker`:

 ```
 nginx: [warn] the "ssl" directive is deprecated, use the "listen ... ssl" directive instead in /etc/nginx/conf.d/mattermost.conf:15
```

**Solution**: The `ssl on;` line should be removed here:

https://github.com/mattermost/mattermost-docker/blob/dd1ecde9c81d390a834d6425c0f86fbada159f41/web/mattermost-ssl#L13-L15

caugner:
#### Summary

Removes the deprecated nginx directive `ssl` which causes a warning when starting `mattermost-docker`.

A replacement is not necessary, because the `ssl` flag is already set on the `listen` directive here:

https://github.com/mattermost/mattermost-docker/blob/492ecd5ca3c761e309834ba0389dd13e7bc431d4/web/mattermost-ssl#L13

#### Ticket Link
Fixes #451.

### DIFF
--- a/web/mattermost-ssl
+++ b/web/mattermost-ssl
@@ -12,7 +12,6 @@ map $http_x_forwarded_proto $proxy_x_forwarded_proto {
 server {
     listen 443 ssl http2;
 
-    ssl on;
     ssl_certificate /cert/cert.pem;
     ssl_certificate_key /cert/key-no-password.pem;
     ssl_session_timeout 5m;


jasonblais:
@DSchalla Reminder to help with review :)

Issue #469, opened 2020-04-21
DB Dockerfile : building failed due to pip version


pjenicot:
Hi,

When I try to build the db image (release 5.22.0), there is an error on the version of PIP :

```
OK: 272 MiB in 80 packages
Collecting wal-e<1.0.0
  Downloading https://files.pythonhosted.org/packages/7a/67/4da340652303e3adb26f6e9aa97b2118dbe5cd17be87e8a21dff9e9bc507/wal-e-0.9.2.tar.gz (100kB)
Collecting envdir
  Downloading https://files.pythonhosted.org/packages/75/dd/817dc77edf4807fc2c29e28046ab83aa7ccfe27e79d24ca37d7aaa1e36bb/envdir-1.0.1-py2.py3-none-any.whl
Collecting gevent>=1.0.2 (from wal-e<1.0.0)
  Downloading https://files.pythonhosted.org/packages/5a/79/2c63d385d017b5dd7d70983a463dfd25befae70c824fedb857df6e72eff2/gevent-1.5.0.tar.gz (5.3MB)
  Could not find a version that satisfies the requirement cffi>=1.12.3 (from versions: )
No matching distribution found for cffi>=1.12.3
You are using pip version 10.0.1, however version 20.1b1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
ERROR: Service 'db' failed to build: The command '/bin/sh -c apk add --no-cache       build-base       curl       libc6-compat       libffi-dev
      linux-headers       python-dev       py-pip       py-cryptography       pv       libressl-dev     && pip --no-cache-dir install -c pip-constraints.txt 'wal-e<1.0.0' envdir     && rm -rf /var/cache/apk/* /tmp/* /var/tmp/*' returned a non-zero code: 1
```
Thank you for your help

zwnk:
i get the same error.

PR #471, opened 2020-04-23
[GH-469] Fixing docker-compose build process

oguera:
#### Summary

Currently the command `docker-compose build` throws an error for building the database because it uses an old pip version.

This PR updates pip right before it is called. Anyways this should just be a temporary fix because the image uses Python 2.7 wich out of support since Jan 2020

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-docker/issues/469


oguera:
I just submitted my CLA

PR #488, opened 2020-08-10
add TLS 1.3 support to web container

nvjacobo:
#### Summary

Add TLS 1.3 support to nginx web server. 

Thanks, Jacobo

### DIFF
--- a/web/mattermost-ssl
+++ b/web/mattermost-ssl
@@ -15,7 +15,7 @@ server {
     ssl_certificate /cert/cert.pem;
     ssl_certificate_key /cert/key-no-password.pem;
     ssl_session_timeout 5m;
-    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+    ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3;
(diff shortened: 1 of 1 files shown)


enahum:
FYI iOS mobile app currently does not support connecting the WebSocket with TLS 1.3

nvjacobo:
> FYI iOS mobile app currently does not support connecting the WebSocket with TLS 1.3

Thanks for your feedback @enahum!

 It seems that iOS mobile app can use the older versions. Support for older versions is not removed.

enahum:
@nvjacobo maybe listed as a known issue in the readme?

nvjacobo:
> @nvjacobo maybe listed as a known issue in the readme?

This change has TLSv1 TLSv1.1 TLSv1.2 TLSv1.3 for chose by client. 

I am reading apple documents (https://developer.apple.com/security/) It seems iOS has support 1.2.

enahum:
IOS does support it, but the WebSocket implementation that we have does not. I know this for a fact as I’ve been running tests for a new implementation because of this issue

enahum:
As long as TLS 1.1 or 1.2 are also enabled, it should be ok

PR #507, opened 2020-12-09
move docker-compose file to an example folder

Aschenbecher:
#### Summary
to prevent docker-compose file got overwritten with every pull

### DIFF
--- a/examples/jwilder/docker-compose.yml
+++ b/examples/jwilder/docker-compose.yml
@@ -1,0 +1,65 @@
+#### you have to copy this file into the root folder of the project !!!
+
+version: "3"
+
+services:
+
+  db:
+    build: db
+    restart: unless-stopped
+    volumes:
+      - .volumes/db/var/lib/postgresql/data:/var/lib/postgresql/data
+      - /etc/localtime:/etc/localtime:ro
+    networks:
+            - internal    
+    environment:
+      - POSTGRES_USER=mmuser
+      - POSTGRES_PASSWORD=mmuser_password
+      - POSTGRES_DB=mattermost
+    # uncomment the following to enable backup
+    #  - AWS_ACCESS_KEY_ID=XXXX
+    #  - AWS_SECRET_ACCESS_KEY=XXXX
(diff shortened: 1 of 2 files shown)


jfrerich:
@angeloskyratzakos and @stylianosrigas, gentle ping for reviews

metanerd (changes requested):
Hi @Aschenbecher , thank you for your PR! I really like the approach to move the docker-compose to a `.example` folder. For contributions you also unfortunately need to sign the CLA still. 
Maybe it would make sense to go with a similar examples approach like nextcloud https://github.com/nextcloud/docker/tree/master/.examples? WDYT?

Aschenbecher:
Hey,

sorry for the late reply.
To be honest, I didnt put much afford into this PR, I was just annoyed, that with every pull the docker compose is overwritten. We had this discussion about this topic already one year ago.
There was a maintainer who was very committed to changing this. But somehow at some point someone changed it back, and now again, the docker compose file is overwritten all the time and I have to make a backup and write it back. It's just a small thing, but it's annoying and the fix is so easy, because of a lot of other repositorys are doing it like this: put the docker-compose files into an examples folder. The only problem is the build step, because the build files have to be relative to the docker-compose files, i didn't found a better solution then to copy the example to the root. Maybe someone else will find a better solution, so maybe you should see this PR more like a suggestion ;)

jaydeland:
Why is this file in a jwilder sub directory?
examples/jwilder/docker-compose.yml

Aschenbecher:
@jaydeland 
Jwilder is a popular reverse proxy. Maybe someone else will add a compose file for traeffik in another folder.

jaydeland (approved):
> @jaydeland
> Jwilder is a popular reverse proxy. Maybe someone else will add a compose file for traeffik in another folder.

oh, I see - sorry for the confusion, I thought this was more generic.

emilyacook:
@Aschenbecher thank you so much for this PR! Would you be open to signing the CLA? It's the last step before we can merge :)

Issue #489, opened 2020-08-15
PostgreSQL 9.6 deprecated


tuxity:
MM 5.26 now supports PostgreSQL 10+, since 9.6 is now deprecated.

https://docs.mattermost.com/administration/changelog.html#compatibility

We need to update https://github.com/mattermost/mattermost-docker/blob/master/db/Dockerfile#L1 to tag `:10-alpine` and provide a documentation to correctly upgrade database with `pg_upgrade`

HarriLuxus:
Has someone tested how to do this?
I'm just wondering would it be enough to do this on running instance (does someone have time to test?)
1. add `RUN pg_upgrade` to db `entrypoint.sh`
2. change the base image in the Dockerfile
3. run `docker-compose build'
4. run `dcoker-compose up -d`
5. remove `RUN pg_upgrade` in db `entrypoint.sh`

tuxity:
Maybe it can be useful https://github.com/tianon/docker-postgres-upgrade

HarriLuxus:
Actually this is still using postgres 9.4
https://github.com/mattermost/mattermost-docker/blob/124a8ba340bc0d3be1ea35f4cdde6aa69403f800/db/Dockerfile#L1

ouafnico:
Mattermost 5.30 is supposed to be out on December. I guess the version of postgresql on this repo will change in the same time?
Has someone tried before?

haivala:
Anyone else find this alarming?

tuxity:
I guess the best would to handle db upgrade in entrypoint.sh script

- Update FROM to postgres:10-alpine
- check in entrypoint.sh (or separated file) the content of PG_VERSION file in postgres data folder
- if 9.4 (or 9.x?) run pg_upgrade command
- startup as normal

can be done manually but would be nice to have it scripted to avoid any user errors

maxraab:
Does this work? Why is no maintainer of this repo caring about this topic?

If we would check the `PG_VERSION` file of the postgres data folder - how could we determine if we're on the next major version? How could we retrive the current version of the database itself? Couldn't we simply do a `pg_upgrade` with every start?

GuidoDr:
well the topic of this thread is not really correct, it is not the 9.6 which is deprecated but the 9.4 that the docker setup of mattermost currently still does use. ;-)

Since I have not seen any progress on this topic so far, I did today, after I had upgraded Mattermost from 5.29.1 to 5.30.1, a manual upgrade of the database.

[...]

PR #523, opened 2021-02-28
Known issue about 5.32 version

nvjacobo:
I added info about issue with 5.32 version. 

Thanks, Jacobo

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ docker stack deploy -c contrib/swarm/docker-stack.yml mattermost
 * Do not modify the Listen Address in Service Settings.
 * Rarely `app` container fails to start because of "connection refused" to
   database. Workaround: Restart the container.
+* Don't update to release-5.32 though, it's broken: #489
 
 ## More information
 


README.md
```diff
@@ -224,6 +224,7 @@ docker stack deploy -c contrib/swarm/docker-stack.yml mattermost
 * Do not modify the Listen Address in Service Settings.
 * Rarely `app` container fails to start because of "connection refused" to
   database. Workaround: Restart the container.
+* Don't update to release-5.32 though, it's broken: #489
```
amyblais, line 227: Would something like this work?

```suggestion
* Upgrading to ``release-5.32`` fails because PostgreSQL 10+ is required. We don't currently have a timeline for the fix, and as an intermediate option we recommend trying out [Mattermost Omnibus](https://docs.mattermost.com/install/mattermost-omnibus.html) which supports PostgreSQL 10+.
```

nvjacobo: I think it is much better and very clear.

amyblais: Thanks, I don't have access to push the edits here.

amyblais (approved):

jasonblais:
@metanerd please help review when you have time

@emilyacook @amyblais should we announce in forums or elsewhere so that the community is aware?

README.md
amyblais: @nvjacobo For those who don't want to use Omnibus (and we don't have a guide to migrate from docker to Omnibus), maybe we could also add a link to this script as a second option (which was created by our support team): https://github.com/mattermost/mattermost-docker/issues/489#issuecomment-790277661.

coltoneshaw: I updated the script a bit to make it easier to use. It's a temporary workaround here.

@tuxity & @GuidoDr outlined the process of how best to do this, I merely scripted it!

PR #527, opened 2021-03-26
Change healthcheck port Mattermost-app

hpwjnijs:
Default MM port is 8065, so lets have the healthcheck at that port


hpwjnijs:
Seems not to be a solution to  'healtcheck' problems.

PR #536, opened 2021-05-24
Support in-database configuration

paddatrapper:
#### Summary
Adds a flag to support in-database configuration through the `CONFIG_USE_DB` environment variable. If this is `true`, the configuration file will not be created if it does not exist. It supports using the custom `MM_SQLSETTINGS_DATASOURCE` if that exists.

On new deployments with `CONFIG_USE_DB` enabled, none of the defaults written into the configuration file are created. Further, it assumes existing installations migrate their configuration over before enabling `CONFIG_USE_DB` as described in the [documentation](https://docs.mattermost.com/administration/config-in-database.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,9 @@ PostgreSQL allows two other variables `sslrootcert` and `sslcrl` for connection 
 If you use a Mattermost configuration file on a different location than the default one (`/mattermost/config/config.json`) :
 * `MM_CONFIG`: configuration file location inside the container.
 
+You can use [in-database configuration](https://docs.mattermost.com/administration/config-in-database.html) by defining:
+* `CONFIG_USE_DB`: `true` to read and store configuration within the database
+
 If you choose to use MySQL instead of PostgreSQL, you should set a different datasource and SQL driver :
 * `DB_PORT_NUMBER` : `3306`
 * `MM_SQLSETTINGS_DRIVERNAME` : `mysql`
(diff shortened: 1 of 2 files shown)


paddatrapper:
Any update here?

jasonblais:
@paddatrapper Sorry for missing your contribution until now.

We are in fact making some changes to the Mattermost Docker repository, and moving to a new project available at https://github.com/mattermost/docker. You can find additional context behind this move here: https://github.com/mattermost/mattermost-docker/issues/548

With that said, would you be open to contributing your changes to the new repository at mattermost/docker instead, and we can have your changes reviewed? 

Sorry for additional work for you. Alternatively we can also open an issue in mattermost/docker to have your changes here migrated to the new repository.

paddatrapper:
There is no entrypoint or docker image builds in https://github.com/mattermost/docker... So this patch will not fit there. It isn't clear where those images are built from. According to Docker Hub, it is [here](https://github.com/mattermost/mattermost-server/blob/master/build/entrypoint.sh), but that has none of the config creation logic, so it appears that the new images do not create any config?

mrckndt:
@paddatrapper the Mattermost binary itself creates a configuration or reads in an existing one. In fact the https://github.com/mattermost/docker setup supports configuration in database. Just export `MM_CONFIG=${MM_SQLSETTINGS_DATASOURCE}` in the enviromnent variables. There is no need to override the entrypoint.

paddatrapper:
@mrckndt so the new images do not require this patch?

mrckndt:
No, and this wasn't needed here too. It should have work by exporting `MM_CONFIG` with the datasource too.

paddatrapper:
> No, and this wasn't needed here too. It should have work by exporting MM_CONFIG with the datasource too.

From experience, it did not work here. That attempted to create the config file the datasource as the filepath, inevitably failing

